PR #9, opened 2021-01-12
Use Ppxlib and remove vendored code

pascutto:


mariojppereira:
I have been able to port the `why3gospel` plugin to this new implementation of `gospel`. Quick question: maybe I missed it, but is this branch already using the pre-processor for GOSPEL comments, in order to avoid shipping a patched version of the OCaml parser?

I am currently working on porting `cameleer`. Do you think we can wait until I am done with it, before merging?

PR #10, opened 2021-01-28
Add workflows to build and publish the doc

pascutto:


pascutto:
Apparently GH actions don't trigger on the PR that set them up; so closing this and iterating on master instead.

PR #81, opened 2021-10-18
Documentation tweaks

jmid:
Here are a few documentation tweaks:
- Most of it is just of the language. I'm not a native English speaker myself so feel free to use what you want of these.
- I found one typo in `first-spec.md` causing the example not to type-check. 
  Using https://github.com/realworldocaml/mdx could prevent such surprises in the future.
- Finally I added a copy of the complete `container.mli` example in `first-spec.md` for convenience.
  as I couldn't find it anywhere. As a reader, at this point I had to copy-paste the different bits together myself.
  Feel free to ignore this commit if you prefer the documentation as is.

### DIFF
--- a/docs/docs/getting-started/first-spec.md
+++ b/docs/docs/getting-started/first-spec.md
@@ -101,7 +101,7 @@ val create: int -> t
 
 Now on to `is_empty` and `mem`.
 
-`is_empty t` returns true if and only if `t` is empty; this is a post-condition. This
+`is_empty t` is true if and only if `t` is empty; this is a post-condition. This
 function also (hopefully) has no side-effect: it does not modify `t`, does not
 depend on any internal state, and does not raise exceptions. In Gospel's
 language, this function is *pure*.


docs/docs/getting-started/first-spec.md
```diff
@@ -101,7 +101,7 @@ val create: int -> t
 
 Now on to `is_empty` and `mem`.
 
-`is_empty t` is true if and only if `t` is empty; this is a post-condition. This
+`is_empty t` returns true if and only if `t` is empty; this is a post-condition. This
```
pascutto, line 104: `is_empty t` does not return `true`, it _is_ `true` :) Can you revert just that bit?

jmid: Hm. I agree that it is standard terminology to say a *function* returns some value. However it is not unusual to say that a *function call* returns some value, see, e.g., 
https://docs.microsoft.com/en-us/dotnet/fsharp/introduction-to-functional-programming/first-class-functions#return-the-value-from-a-function-call  for someone else in FP also using this terminology.
I guess it depends on whether you consider the call as completed or not. Operationally, when the call hasn't been carried out or hasn't completed yet there is no return value to talk of. For an application in logic this is of course a different matter...
Anyway, I reverted the change as you already saw.

backtracking:
Yes, thanks @jmid !

Issue #39, opened 2021-09-06
Issue a warning when `old` is used on a read-only expression


pascutto:
`old e` is effectively just `e` when `e` is read-only, so the `old` primitive is likely an error from the programmer; we want to issue a warning if that happens.

Issue #37, opened 2021-09-01
Issue a warning when a function is `... -> unit` but has no `modifies` clause


pascutto:
Most of the time, functions returning `unit` have side effects.
If the side effect affects some observable data, then there should be a `modifies` clause.
If it is another kind of side effect, the function probably can't be specified anyway

PR #123, opened 2021-12-17
Warnings

n-osborne:
Concerns issues #37 and #39

Warnings are printed on stderr.


src/typing.ml
```diff
@@ -829,6 +891,14 @@ let process_val ~loc ?(ghost = Nonghost) kid crcm ns vd =
   in
   let spec = process_val_spec kid crcm ns id args ret spec in
   let so = Option.map (fun _ -> spec) vd.vspec in
+  let () =
+    (* check there is a modifies clause if the return type is unit, through a warning if not *)
+    if Ttypes.(ty_equal ret { ty_node = Tyapp (ts_unit, []) }) then
```
pascutto, line 896:
```suggestion
    if Ttypes.(ty_equal ret ty_unit) then
```

src/typing.ml
```diff
@@ -635,6 +636,60 @@ let rec val_parse_core_type ns cty =
...
+  let rec check t =
+    match t.t_node with
+    | Tvar vs ->
+        if
+          not
+            (List.exists
+               (function { t_node = Tvar vs' } -> vs = vs' | _ -> false)
+               wr)
+        then Some vs.vs_name.id_str
+        else None
```
pascutto, line 650: I don't think that would be enough. For instance, you would reject the following:
```ocaml
val f : 'a array -> unit
(*@ f x 
    modifies x 
    ensures let y = x in old y.(0) = 0
```
You need a more subtle data flow analysis to know which variables are actually modified when you write `modifies x`.

src/warnings.ml
```diff
@@ -0,0 +1,29 @@
...
+type kind =
+  | Old_on_read_only of string
+  | Return_unit_without_modifies of string
+
+let pp_kind ppf = function
+  | Old_on_read_only var ->
+      pf ppf "There is a reference to old %s, but is it read only." var
+  | Return_unit_without_modifies fct ->
+      pf ppf
+        "The function %s returns a unit but its specifications does not \
```
pascutto, line 13:
```suggestion
        "The function `%s' returns a unit but its specifications does not \
```

src/warnings.ml
```diff
@@ -0,0 +1,29 @@
...
+         contains any modifies clause."
```
pascutto, line 14:
```suggestion
         contains any `modifies' clause."
```

pascutto (commented):
Thanks ! I would argue for splitting the two new warnings, since one seems easier to implement than the other

n-osborne:
> Thanks ! I would argue for splitting the two new warnings, since one seems easier to implement than the other

You mean in two different PRs?

PR #139, opened 2022-02-03
remove dead code

n-osborne:
- type `ghost` in `uast.mli`, so that merlin does not call every bool a ghost.
- field `p_vars` of pattern (never accessed except for constructing pattern from patterns).


pascutto:
@mariojppereira is it fine with you that `p_vars` is removed (wrt your usage in `cameleer`)?

PR #165, opened 2022-03-18
Add a new `equality` clause

pascutto:
This adds a new clause `equality`, which lets the user specify that a function `val equal : t -> t -> bool` is an equality function over the type `t`.


mariojppereira:
I am really curious: what is the use case for this `equality` functions? Do you have an example in Ortac?

PR #172, opened 2022-03-30
PatternBadtype: fix order in error message

paulpatault:


pascutto (approved):
It does seem right indeed, thanks!

PR #173, opened 2022-03-31
Ppxlib.0.26.0 compatibility

pitag-ha:
This is a patch PR to make the PPX compatible with ppxlib.0.26.0 which has bumped the AST to 4.14/5.00.

With this patch, the ppx just errors out when one of the new features is encountered in the AST. That's because I myself don't know all >100 ppxlib reverse dependencies well enough to decide what the behavior should be when encountering the new feature. It would be great though if you, the maintainers, could follow up on this and implement the correct behavior. For example, I'd guess that in `Opprintast`, `constructor_declaration` should be adapted to take into account the new `vars` field that captures explicit binders for type variables.

The two new features I'm talking about are naming GADT existential (see [this section](https://ocaml.org/manual/gadts-tutorial.html#s:explicit-existential-name) in the manual or PR description [here](https://github.com/ocaml/ocaml/pull/9584)); and explicit binders for type variables (see PR description [here](https://github.com/ocaml/ocaml/pull/10437)).

### DIFF
--- a/gospel.opam
+++ b/gospel.opam
@@ -24,7 +24,7 @@ depends: [
   "cmdliner" {>= "1.1.0"}
   "fmt" {>= "0.8.7"}
   "ocaml-compiler-libs" {>= "v0.12.0"}
-  "ppxlib" {>= "0.23.0"}
+  "ppxlib" {>= "0.26.0"}
   "ppx_deriving" {>= "5.2.1"}
 ]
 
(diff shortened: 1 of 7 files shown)


pascutto:
Awesome, thanks Sonja! 
I will push our implementation of the newly added bits on top of this PR, then merge it.

PR #182, opened 2022-04-11
update CHANGES

n-osborne:
Since we havent't done that regularly, this PR update the
`CHANGES.md` file with the change since last realease.

I may have miss some, feel free to add them and/or modify the presentation.


pascutto:
Thanks! Updated with more entries, and added a CI check so we don't forget in the future!

PR #183, opened 2022-04-11
Type variable in ts_tuple and fs_tuple should be consistent

paulpatault:


pascutto (approved):
Thanks! Was there an occurrence where this bug was visible from the outside?

PR #187, opened 2022-04-20
Add `self` in type invariants

pascutto:
Users can now refer to `self` in invariants, which refers to the value of the type the invariant applies to. This lets us specify invariants on recursive types, and pattern matching on variant types in invariants.

Maybe a more debatable point: this PR also removes the ability to refer to a type model or field by just referring to the field, now we have to use `self` all the time. This is not necessary, but I figured that would push for uniform specs, and gets closer to OCaml, although I agree it makes the invariants a little bit more verbose.
To make thinks clearer, 
```ocaml
type t = {x : int; y : int}
(*@ invariant x > 0 *)
```
now has to be written
```ocaml
type t = {x : int; y : int}
(*@ invariant self.x > 0 *)
```

 Happy to hear thoughts about that!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Added
 
+- Added a `self` variable in type invariants referring to a value of the
+  specified type. [\#187](https://github.com/ocaml-gospel/gospel/pull/187)
 - Added support for `int` literals.
   [\#175](https://github.com/ocaml-gospel/gospel/pull/175) and
   [\#177](https://github.com/ocaml-gospel/gospel/pull/177)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,8 +2,9 @@
 
 ## Added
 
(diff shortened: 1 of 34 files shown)


fpottier:
I think it is a good move to write `self.x` instead of just `x`. This seems more general, because not every type is a record type. E.g., if `t` is a sum type, then the invariant is likely to be of the form `match self with ...`.

I note that we might wish to debate whether it is good idea to bind the name `self` by default. In Gospel function specifications, we do not bind any name by default, and we require the user to explicitly name every argument and result. For consistency, we might wish to let the user choose the name of `self` in invariants. E.g., we might write
```
(*@ invariant this : t implies this.x > 0 *)
```
This is admittedly somewhat verbose and redundant, but I thought I would raise this point, because it seems a little inconsistent to sometimes use explicit binding and sometimes use fixed, implicitly-bound names.

PR #189, opened 2022-04-20
Better error handling

pascutto:
This refactors the error handling to remove all the exceptions (and exception printers registrations), which were not suitable for a library usage of Gospel.
All errors now raise an `Error of loc * kind`, which is easy to catch and print by client libraries


src/Opprintast.ml
```diff
@@ -11,7 +11,7 @@
 (*                                                                        *)
 (*   All rights reserved.  This file is distributed under the terms of    *)
 (*   the GNU Lesser General Public License version 2.1, with the          *)
-(*   special exception on linking described in the file LICENSE.          *)
+(*   special exception on linking described in the file LICENSW.          *)
```
Armael, line 14: erroneous change?

src/coercion.ml
```diff
@@ -5,14 +5,13 @@
 (*                                                                  *)
 (*  This software is distributed under the terms of the GNU Lesser  *)
 (*  General Public License version 2.1, with the special exception  *)
-(*  on linking described in file LICENSE.                           *)
+(*  on linking described in file LICENSW.                           *)
```
Armael, line 8: same

src/parser_frontend.mli
```diff
@@ -5,37 +5,34 @@
 (*                                                                  *)
 (*  This software is distributed under the terms of the GNU Lesser  *)
 (*  General Public License version 2.1, with the special exception  *)
-(*  on linking described in file LICENSE.                           *)
+(*  on linking described in file LICENSW.                           *)
```
Armael, line 8: same

Armael (commented):
(small refactoring bug)

src/Opprintast.ml
pascutto: Ah yes, my `sed` was a bit too brutal. Thanks for catching this!

PR #211, opened 2022-05-30
Always add constants to the logical namespace

### LINKED ISSUE #25: modifies and global variables
The following is currently rejected:
```
val a : int ref
val f : int ->  int
(*@ y = f x modifies a *)
```

pascutto:
Fixes #25.
Constants can now be referenced in specifications.

Note: a constant here is a symbol that doesn't take arguments. This is not properly computed at the moment by Gospel in the presence of alias types.
This can be fixed separately as it goes beyond #25.


pascutto:
The CI fails because the stdlib in OCaml `< 4.12` did not expose `__FUNCTION__`, so the printing of the namespace is different.
I think this is (another?) argument for removing printing-based regression tests. Those were useful in the beginning but are not very practical anymore now.

Issue #202, opened 2022-05-27
Improve pattern-matchings


pascutto:
- [ ] Add a `when` guard in pattern cases.
- [ ] Check for pattern redundancy and provide example.
- [ ] Check for ambiguous patterns (see https://github.com/ocaml-gospel/gospel/issues/192)

PR #213, opened 2022-05-31
Pattern redundancy

paulpatault:
Related to #202 
This performs a redundancy check for pattern-matching.

### DIFF
--- a/src/patmat.ml
+++ b/src/patmat.ml
@@ -31,6 +31,7 @@ let mex a =
     n
   with Brk i -> i
 
+(* To be removed when upgrade to OCaml 4.11 will be done *)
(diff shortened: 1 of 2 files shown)


src/patmat.ml
```diff
@@ -31,6 +31,13 @@ let mex a =
     n
   with Brk i -> i
 
+let filteri p l =
```
pascutto, line 34: Can you add a comment above so we don't forget to remove this when we upgrade to 4.11?

pascutto (approved):
LGTM, thanks!
Don't forget the changelog entry here aswell, perhaps in the same entry as the exhaustivity one.

PR #214, opened 2022-05-31
Highlight source locations when reporting errors

Armael:
This PR adds support for reporting source location when displaying errors, similar to the error reports in OCaml >= 4.08. This relies on the [pp_loc](https://github.com/Armael/pp_loc) library.
The easy part was plugging in the library. 
The harder part was to fix the locations produced by gospel. The solution was twofold:
- modify the preprocessor to carefully preserve column alignment of the payload within `(*@ *)` blocs (by emitting newlines within  the `[@@@gospel ...]` where appropriate), using line directives to make up for the extra lines introduced
- carefully handle and "repair" the locations at the time of displaying the error, to only rely on line and column numbers, but not byte offsets from the start of the file (which happen to refer to the file after preprocessing rather than before preprocessing as one might hope).

Most of the tricky location handling logic is factored out as helpers in pp_loc, so here most of the code is about modifying the preprocessor.

The result of the new error messages can be seen in the last commit, which updates the .expected files of the testsuite.


pascutto (approved):
This is really great, thank you!
I'm surprised (in a good way) that the locations were not _too_ broken besides the preprocessing part.
Could you also add a changelog entry under "Improved"?
You're also missing a `dune build @fmt --auto-promote`.
And `Lexing.set_filename` is not supported before 4.11 unfortunately.

pascutto:
I just pushed a change in the test format to allow this. Expected results are now displayed in 
```
(* {gospel_expected|
   ...
   |gospel_expected} *)
```
Let me know what you think @Armael :)

pascutto:
It looks like `pp_loc` outputs a different message depending on the OCaml version (with a change in 4.11). Is that a known behaviour?
The locations seem wrong on < 4.11

Armael:
ah, I see. Upgrading to 4.11 sounds good to me.

PR #220, opened 2022-06-02
Added full guard pattern error

paulpatault:
Add specific error message for patterns with guard on every clause.


pascutto (changes requested):
Thanks! Just a very minor fix :)

pascutto (approved):
Thank you!

PR #222, opened 2022-06-05
Change error to warning for redondant patterns

paulpatault:


pascutto:
Thanks!
I'm not sure it is a good idea that the `gospel` library prints error messages at all, I would argue for something that lets its user decide how to handle the case where an error/warning happens, otherwise it becomes hard to detect these from the client code.

Issue #194, opened 2022-05-12
integer literal in hexadecimal not accepted


GoncaloMateus:
Error Message 

"Invalid integer literal in base 10: '0x16' "

PR #223, opened 2022-06-05
Hexadecimal base int

paulpatault:
Added test relative to past error #194


pascutto (approved):
LGTM, thanks! Can you keep only the last commit in this PR so we can handle the warnings separately?

paulpatault:
> LGTM, thanks! Can you keep only the last commit in this PR so we can handle the warnings separately?

Done 👍

PR #228, opened 2022-07-10
Changed modifies clause warning to exclude values of type unit

naomiiiiiiiii:
An error is thrown when values of type unit are not specified with a 'modifies' clause.  

### Example
With file
```
(*example.mli*)
val x : unit
(*@*)
```
running `ortac example.mli` produces
```
~ ᐅ ortac example.mli
File "example.mli", line 1, characters 0-18:
1 | val x : unit
2 | (*@*)
Error: The function `x' returns `unit'
       but its specifications does not contain any `modifies' clause.
```
### Cause 
The error is caused by lines 903-909 in `gospel/src/typing.ml`.  I think the intent of this code (from #37 ) is to ensure that all functions which return unit are specified as modifying something. However, as it stands, values of type unit are also required to modify something. I am not sure if this is a design decision. 

### Solution
To allow values of type unit to be pure, I added a check on line 904. I also added a test in 
`gospel/test/warnings`, next to the test for functions with return type unit.

### DIFF
--- a/src/typing.ml
+++ b/src/typing.ml
@@ -900,8 +900,8 @@ let process_val ~loc ?(ghost = Nonghost) kid crcm ns vd =
   let spec = process_val_spec kid crcm ns id args ret spec in
   let so = Option.map (fun _ -> spec) vd.vspec in
   let () =
-    (* check there is a modifies clause if the return type is unit, through a warning if not *)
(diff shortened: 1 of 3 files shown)


pascutto (approved):
Thanks! (and sorry again for the delay)

PR #253, opened 2023-01-30
Fix specifications in markdown documentation

shym:
Add a few `Makefile` rules to check the Gospel specifications inserted as OCaml code blocks by extracting the code blocks to a corresponding *.mli file and calling `gospel check` on it
Skip the documentation files intentionally using invalid syntax or citing code blocks that cannot all be exported to one file
Add the generated *.mli files to .gitignore

This reveals issues in quite a few documentation files

### DIFF
--- a/docs/docs/dune
+++ b/docs/docs/dune
@@ -1,0 +1,3 @@
+(include ../../test/utils/dune.docs)
+
+(include dune.inc)
--- a/docs/docs/dune.inc
+++ b/docs/docs/dune.inc
@@ -1,0 +1,42 @@
+(rule
+ (enabled_if %{bin-available:awk})
+ (deps
+  (:md2mli %{project_root}/test/utils/md2mli.awk))
+ (action
+  (with-stdout-to
+   good-practices.mli
+   (chdir %{project_root}
+    (run awk -f %{md2mli} %{dep:good-practices.md})))))
+
+(rule
+ (alias runtest)
+ (enabled_if %{bin-available:awk})
+ (deps %{bin:gospel})
+ (action
+  (progn
+   (with-stdout-to good-practices.output
+    (chdir %{project_root}
+     (run gospel check %{dep:good-practices.mli})))
+   (diff? %{project_root}/test/utils/check_success good-practices.output))))
+
+(rule
+ (enabled_if %{bin-available:awk})
+ (deps
+  (:md2mli %{project_root}/test/utils/md2mli.awk))
+ (action
+  (with-stdout-to
+   welcome.mli
+   (chdir %{project_root}
+    (run awk -f %{md2mli} %{dep:welcome.md})))))
+
+(rule
+ (alias runtest)
+ (enabled_if %{bin-available:awk})
+ (deps %{bin:gospel})
+ (action
+  (progn
+   (with-stdout-to welcome.output
+    (chdir %{project_root}
+     (run gospel check %{dep:welcome.mli})))
+   (diff? %{project_root}/test/utils/check_success welcome.output))))
+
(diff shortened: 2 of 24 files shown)


shym:
I’m not sure how to fix the syntax errors on `ephemeral`.

Makefile
```diff
@@ -7,10 +7,38 @@ install:
...
-	dune clean
-
 format:
 	dune build @fmt --auto-promote
 
-.PHONY: all gospel clean format test
+
+# To check that the markdown documentation is kept valid
+
+GOSPEL_CMD := dune exec --display=quiet -- bin/cli.exe
```
pascutto, line 16:
```suggestion
GOSPEL_CMD := dune exec --display=quiet -- gospel
```

Makefile
```diff
@@ -7,10 +7,38 @@ install:
...
+# DOCUMENTS += docs/docs/walkthroughs/fibonacci.mli # `let` in OCaml block
+DOCUMENTS += docs/docs/walkthroughs/mutable-queue.mli
+DOCUMENTS += docs/docs/walkthroughs/union-find.mli
+# DOCUMENTS += docs/src/pages/faq.mli # `...` in OCaml block
+CHECKS := $(patsubst %,%.check,$(DOCUMENTS))
+
+.PRECIOUS: $(DOCUMENTS)
+
+%.mli: %.md
+	awk '/^ *```ocaml/,/^ *```$$/ { if ($$0 !~ "^ *```") print }' < $< > $@
```
pascutto, line 33: I'm really bad at this so I won't try to fix it, but
```
val f : t
(*@ ... *)

val g : t
(*@ ... *)
```
seems to be rewritten as 
```
val f : t

(*@ ... *)
val g : t

(*@ ... *)
```
which could (should) be an incorrect attachment of the specs.

pascutto (commented):
Thanks!
Do you think you could make this happen as part of the docs CI?
Also, is there a reason to do it using a `Makefile` rather than a dune rule plugged to the tests? That would also avoid polluting the workspace with the temporary test files.

About the `ephemeral` issue, this is a bug in the preprocessor, not attaching the spec attribute to the string in the floating `[@@@gospel "type t"]` attribute because of the interleaved comment.
The preprocessor doesn't know (yet) about comments or docstrings, so it just treats the comment as a string it shouldn't touch (usually OCaml code) and tries to attach the spec to it, which is indeed a syntax error.

shym:
Quasi-complete re-roll of that PR:
- using `dune` rules instead of Makefile (as a side-note, I find it a lot more inefficient (and verbose!) when fixing the syntax errors, because errors are reported against the `.mli` file that is hidden away somewhere, I cannot use standard tooling to go to the location and see the error in more context than what `gospel check` reports)
- generate .mli with skipping lines, to avoid ambiguities,
- fixing a large part of the syntax errors and annotating code blocks that are not using valid syntax.
This triggers 2 `gospel` crashes that I didn’t investigate.

shym:
A large update of that PR to prepare its merge for the upcoming release:

- rebased on top of #312, to store also in `test/utils` the code that tests the code extracts in markdown documentation (without creating unduly conflicts later), so that this PR is really only the last 4 commits,
- used the location information so that syntax errors are reported to the markdown file when possible (`gospel pps` breaks that, maybe it could pick up `# line` data to update its location accordingly?),
- fixed up a couple of other syntax errors I found,
- commented on the code extracts that are tagged as _invalid_ with the reason why, to help fix those discrepancies between the documentation and `gospel check` (in particular on `ephemeral`).

shym:
Rebased on `main`, now that #312 has been merged.

n-osborne (approved):
Nice automation that already proven its worth by all the correction!
Thanks!

shym:
Approved by human & CI, and it is used as the starting point of another PR, so I merge it now.

PR #262, opened 2023-02-08
Display an “Internal error” on dummy positions

shym:
Instead of running into an exception, display an ”Internal error” message when a warning is attached to a dummy position

I didn’t investigate why such dummy positions are generated.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,6 +50,9 @@
 
 ## Fixed
 
+- Avoid uncaught exception when displaying a warning on a dummy
+  position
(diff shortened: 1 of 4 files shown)


shym:
I imagine such a change should appear in the changelog, seeing a few others.
Is there a policy somewhere about what should or should not be listed in the changelog?

shym:
After further investigations, only the filename is missing, other position data is available, so I updated the error message accordingly. But I still didn’t find where that partial location is (mis-)built.

n-osborne (approved):

PR #264, opened 2023-02-10
Gospel preprocessing in order to make odoc display specifications as documentation

n-osborne:
This PR proposes another command line subcommand to gospel that turn the gospel comments into documentation comment. Doing so it places the contents of the gospel specifications into an odoc code bloc labeled with `gospel`.

This is done by reusing the existing preprocessor, making the `run` function taking one parameter more that tell it how to print the gospel comments (the original one put them into an OCaml attribute labeld with `gospel`, the new one put them inside documentation markers).

The content of the specifications are preceded by a "Gospel specifications:" header (inside the code bloc, I think it is more readable, but that's not a strong opinion).
Gospel ghost declaration (gospel functions, predicates, axioms...) are preceded by a "Gospel declaration:" header (again inside the code bloc, for the same reason and still not a strong opinion).

I think this begins to give a quiet readable display, as an OCaml value (or type) will be followed by its informal documentation and then the formal documentation from gospel will follow. Then, usually, there will be another OCaml value or type that will separate nicely the parts of the module's documentation (as nicely as it is with just the informal documentation at least).

The only thing the user has to do (beside specifying the library)  is to add a preprocess action in the dune file of the said library.

An example of how to use the proposed new subcommand can be found in [this repo](https://github.com/n-osborne/gospel-odoc-example). To see the result in an HTML page, just run `dune build @doc` after installing the gospel version of this PR.


n-osborne:
These tests reveal that this approach does not play nice with ocamlformat.
I had to add two options:
- `parse-docstring = false`
- `doc-comments = before`

n-osborne:
Related to the limitation of this approach, an `ocamlformat` user should set `doc-comments` to `before`.

n-osborne:
It turns out (#266) that `gospel pps` already skip documentation comments that are below a declaration when there is a specification attached to the latter.

Hence, fixing this limitation here won't have any effect while `gospel pps` keeps this behaviour.

n-osborne:
Converting to draft as I explore another (arguably better) option [here](https://github.com/n-osborne/ppx_odoc_of_gospel)

shym:
Should this feature (whatever the implementation) be considered for the 0.2 release or the next one?

n-osborne:
Superseed by #288

PR #267, opened 2023-03-21
Fix pps with documentation

### LINKED ISSUE #266: gospel pps detaches documentation comments when below and followed by a specification
This behaviour can be reproduced as follow:

```
$ cat > ./bug_pps.mli << EOF
> val f : int -> int
> (** blah *)
> (*@ y = f x *)
> EOF
```
 Then we have:
```
$ ocamlc -dsource bug_pps.mli
val f : int -> int[@@ocaml.doc " blah "]
```
Which is the expected behaviour.

But:
```
$ ocamlc -pp "gospel pps" -dsource bug_pps.mli
val f : int -> int[@@gospel {| y = f x |}]
```
Where the documentation comment has been lost.

n-osborne:
This PR's intent is to fix #266 

A Documentation token is added to the lexer, so that we can print the ones between an OCaml declaration and gospel specifications directly as attributes. The other are printed as documentation comments, so that OCaml parser deal with them appropriately.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -52,6 +52,9 @@
 
 ## Fixed
 
+- Documentation comments between declarations and specifications are not
+  detached anymore
+  [\#266](https://github.com/ocamlc-gospel/gospel/pull/266)
 - Avoid uncaught exception when displaying a warning on a dummy
   position
   [\#262](https://github.com/ocaml-gospel/gospel/pull/262)
--- a/dune-project
+++ b/dune-project
@@ -1,3 +1,4 @@
-(lang dune 2.4)
+(lang dune 2.7)
 (name gospel)
 (using menhir 2.0)
+(cram enable)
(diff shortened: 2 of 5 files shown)


n-osborne:
CI is green-ish (the failure is a git-fetch one).
So PR is ready for review.

pascutto:
Suggestion: maybe the docs preprocessor should apply to OCaml ASTs rather than strings; that looks simpler to implement.

n-osborne:
I'm working on a ppx rewriter that turns the Gospel attributes into documentation attributes ([here for now](https://github.com/n-osborne/ppx_odoc_of_gospel)). But this ppx relies on having the Gospel content in Gospel attributes, that is it relies on `gospel pps`.

The issue I've stumbled on (#266) is that the current gospel preprocessor (`gospel pps`) detaches some of the documentation (if we want the gospel attributes in the OCaml AST, we may lose some of the documentation ones).

This PR proposes a fix to that issue, so that we have all the documentation attributes in the OCaml AST, even after `gospel pps`.

Issue #128, opened 2022-01-06
Ghost variables 


pascutto:
Some ideas on improving the handling of ghost variables in the AST:
- Add a `ghost` field to `vsymbol` instead of having a `Lghost` label
- Change the syntax, e.g. to `[@ghost x: t, y: t']`, to disambiguate with lists and allow patterns as returned values.

PR #271, opened 2023-04-13
Remove Lghost constructor

n-osborne:
This PR address part of #128, namely removing the `Lghost` label and storing the corresponding piece of information elsewhere. In order to keep track of the ghost information, we've added:
- a `ghost` field to `vsymbol`
- a `ghost` argument (holding an optional type information) to `Uast.Lnone`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -87,6 +87,8 @@
 
 ## Internals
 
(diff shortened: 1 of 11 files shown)


n-osborne:
Closing because it is stale.

PR #277, opened 2023-05-12
Fix xpost pattern match

### LINKED ISSUE #237: Allow pattern-matching in raises (to correspond with docs)
In the gospel docs ([here](https://ocaml-gospel.github.io/gospel/language/function-contracts)) there is the following example for exceptional postconditions:
```
(*@ ...
    raises Error "foo" -> P | Error _ -> Q
    raises Error x -> R *)
```
However, with the following file, 

```
(*example.mli*)
exception Error of string

val silly : int -> int
(*@
    raises Error "foo" -> 1 = 1 | Error _ -> 10 = 10
    raises Error x -> 42 = 42 *)
```
I get
```
~ ᐅ gospel check example.mli
File "example.mli", line 5, characters 11-52:
5 |     raises Error "foo" -> 1 = 1 | Error _ -> 10 = 10
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Type checking error: Exception pattern does not match its type.
```
Simpler tests produce the same problem. For example,

```
(*example1.mli*)
exception Error of string

val silly : int -> int
(*@
    raises Error "foo" -> 1 = 1 *)
```
gives
```
~ ᐅ gospel check example1.mli
File "example1.mli", line 5, characters 11-31:
5 |     raises Error "foo" -> 1 = 1 *)
               ^^^^^^^^^^^^^^^^^^^^
Error: Type checking error: Exception pattern does not match its type.
```
and
```
(*example2.mli*)
exception Error of int

val silly : int -> int
(*@
    raises Error 0 -> 1 = 1 *)
```
gives 
```
~ᐅ gospel check example2.mli
File "example2.mli", line 5, characters 11-27:
5 |     raises Error 0 -> 1 = 1 *)
               ^^^^^^^^^^^^^^^^
Error: Type checking error: Exception pattern does not match its type.
```

Only the most general pattern seems to work. For example, 
```
(*example3.mli*)
exception Error of int

val silly : int -> int
(*@
    raises Error z -> 1 = 1 *)
```
gives
```
~ᐅ gospel check example3.mli
OK
```

n-osborne:
This PR fix #237 

Add support for:
- constant pattern (*eg* `E 'a'`)
- intervals (*eg* `E ('a'..'z')`)
- pattern with alias (*eg* `E (x as y)`)
- disjunctive pattern (*eg* `E ('a' | 'b')`)

Note: 
- casting in exceptional pattern in not supported (it is not in `Dterm.pattern` neither)
- pattern constructor `Papp` is not supported neither


shym:
For the record, we took the review offline.

PR #281, opened 2023-05-30
Add a documentation tokens to pps

### LINKED ISSUE #266: gospel pps detaches documentation comments when below and followed by a specification

n-osborne:
This PR proposes a fix for #288

It adds two new tokens to the lexer:
- a `Documentation` token with starting and ending position and the contents
- an `Empty_documentation` token with the ending position

Both are printed as documentation attributes when they appears between an Ocaml declaration and a gospel specification (the case where the documentation comments is considered as detached by the OCaml parser)

Tests are added to check that the documentation is not considered as detached anymore and to make sure that original position is preserved.

### DIFF
--- a/src/pps.mll
+++ b/src/pps.mll
@@ -68,6 +68,32 @@
       end_p.Lexing.pos_lnum end_p.pos_fname
       (String.make (end_p.pos_cnum - end_p.pos_bol - 1 (*]*)) ' ')
 
+  let print_documentation_attribute start_p end_p s =
+    Fmt.str "[@@@ocaml.doc\n\
+# %d \"%s\"\n\
+%s {|%s|}\n\
+# %d \"%s\"\n\
+%s]"
+    start_p.Lexing.pos_lnum start_p.pos_fname
+    (String.make (start_p.pos_cnum - start_p.pos_bol) ' ') s
+    end_p.Lexing.pos_lnum end_p.pos_fname
+    (String.make (end_p.pos_cnum - end_p.pos_bol - 1 (*]*)) ' ')
+
+  let print_empty_documentation end_p =
+    Fmt.str "[@@@ocaml.doc\n\
+# %d \"%s\"\n\
+%s]"
+    end_p.Lexing.pos_lnum end_p.pos_fname
+    (String.make (end_p.pos_cnum - end_p.pos_bol - 1 (*]*)) ' ')
+
+  let print_triplet o sp doc sp' start_p end_p s =
+   let docstring = match doc with
+   | Documentation (start_p, end_p, d) -> (print_documentation_attribute start_p end_p d)
+   | Empty_documentation end_p -> print_empty_documentation end_p
+   | _ -> assert false
+   in
+   Fmt.str "%s%s%s%s%s" o sp  docstring sp' (print_gospel `TwoAt start_p end_p s)
+
   let rec print = function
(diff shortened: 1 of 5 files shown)


src/pps.mll
```diff
@@ -151,6 +170,19 @@ and gospel start_pos = parse
       scan lexbuf
     }
 
+and documentation start_pos = parse
+  | blank+ as s { Buffer.add_string buf s; documentation start_pos lexbuf }
+  | newline as nl { Buffer.add_string buf nl; Lexing.new_line lexbuf; documentation start_pos lexbuf }
+  | "*)" {
```
shym, line 176: This doesn’t support nested comments, but it would be nice to make sure they are, using a similar trick to was is done for regular comments.

test/documentation/test_preprocess_documentation.t
```diff
@@ -0,0 +1,65 @@
+Testing interaction between `gospel pps` and Ocaml compiler. The issue is that
```
shym, line 1: Very nice test!
Doesn’t it miss some dune configuration to be run? I intentionally broke the test and `dune runtest` didn’t complain.

src/pps.mll
```diff
@@ -82,8 +95,12 @@
     Fmt.str "%s%s" (print_gospel `TwoAt start_p end_p s) (print l)
   | Other o :: Spaces sp :: Spec (start_p, end_p, s) :: l ->
     Fmt.str "%s%s%s%s" o sp (print_gospel `TwoAt start_p end_p s) (print l)
+  | Other o :: Spaces sp :: Documentation (doc_start_p, doc_end_p, d) :: Spaces sp' :: Spec (start_p, end_p, s) :: l ->
```
shym, line 98: That would be admittedly rather ugly, but the `Spaces` mean that
```
val f: int -> int(** doc *)(*@ spec *)
```
isn’t covered as one would hope.

shym (commented):
Very nice job!
I think I found two corner cases that it would be nice to cover before merging.

test/documentation/test_preprocess_documentation.t
```diff
@@ -0,0 +1,134 @@
...
```
shym, line 1: The tests could also use an inner nested comment, since they should be now covered.

src/pps.mll
```diff
@@ -120,6 +165,21 @@ rule scan = parse
     { push (); Lexing.new_line lexbuf; Queue.push (Spaces nl) queue; scan lexbuf }
   | "(*@"
     { push (); gospel (Lexing.lexeme_start_p lexbuf) lexbuf }
+  | "(**)" {
+    push ();
+    let end_pos = Lexing.lexeme_end_p lexbuf in
+    Queue.push (Empty_documentation end_pos) queue;
+    scan lexbuf }
```
shym, line 172: The two new rules use different block endings (`}` on a line by itself or not.

shym (approved):
Very nice job!
I managed to find two minor points just to quibble :-) but LGTM.

PR #284, opened 2023-06-26
Accept partial application by using eta-expansion on the fly

shym:
Before, the following expressions were refused:

```
11 |     ensures sum = List.fold_left (+) 0 ns
                                      ^^^
Error: The symbol `infix +' cannot be partially applied.
```

```
11 |     ensures sum = List.fold_left (fun x -> (+) x) 0 ns *)
                                                ^^^
Error: The symbol `infix +' cannot be partially applied.
```

This quick fix solves the issue by an η-expansion on the fly with dummy variable names, so this is not completely satisfactory as the dummy names will spill to users of the library. It breaks the code generated by Ortac, for instance.

### DIFF
--- a/src/typing.ml
+++ b/src/typing.ml
@@ -273,6 +273,27 @@ let rec dterm kid crcm ns denv { term_desc; term_loc = loc } : dterm =
     if ls.ls_field then W.error ~loc (W.Field_application ls.ls_name.id_str);
     gen_app ~loc ls tl
   in
+  let eta_expand q args missing =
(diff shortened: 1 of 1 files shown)


n-osborne (commented):
Thanks!
This is nice to have a quick fix at hand.
I am a bit reluctant to break the code generated by Ortac, and more generally to spill the dummy names created for internal reason to users (as you mention yourself).

I've experimented a bit [here](https://github.com/n-osborne/gospel/tree/experiment-with-bad-arity), but still not close to a better solution.

shym:
Closing, as this has been superseded by #290.

PR #285, opened 2023-06-27
Route location information for type applications

### LINKED ISSUE #258: Bad location for a type error
Given the simple file `ex.mli`:
```
type 'a t
val create: int -> t
```
`gospel check` reports:
```
$ gospel check ex.mli 
File "ex.mli", line 1, characters 8-9:
1 | type 'a t
            ^
Error: The type `t' expects 1 argument(s) but was given 0 argument(s) here.
```

shym:
Add an optional location to `ty_app` using the location of the declaration of the type only when the precise application location is missing
Add tests for type arity errors

Closes #258


n-osborne (commented):
Nice fix.
Need formatting before merging.
But maybe we should upgrade to ocamlformat 0.25.1 and then merge this PR formatted

shym:
> But maybe we should upgrade to ocamlformat 0.25.1 and then merge this PR formatted

I’d like that option, I just opened #286 for that.

shym:
Rebased and reformatted. :crossed_fingers:

PR #288, opened 2023-07-03
A ppx to display gospel contents as documentation with odoc

n-osborne:
This PR proposes to add a ppx rewriter to build documentation attributes from
the contents of the gospel ones.

The documentation attributes are added, they don't replace the gospel ones.

Also, as we are post-parsing there is no problem of having more than two
documentations attributes. That means that if a signature item has already two
documentations attributes from the parsing, building one with the gospel
contents does not raise any problem (especially, odoc will treat all of them)

This PR proposes to name the ppx `gospel.ppx` so the dune file would look like that:

```dune
(library
  (name library_name)
  (preprocess
  (pps  gospel.ppx -- -pp "gospel pps")))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,20 @@ $ gospel check max_array.mli
 OK
 ```
 
+Gospel also provides a ppx rewriter to allow odoc to display the contents of
+gospel specifications and declarations as documentation. This ppx rewriter
+works under the assumption that the source preprocessor has been run first. The
+dune stanza reads as follows:
+
+```dune
+(library
+ (name lib_name)
+ (preprocess
+  (pps gospel.ppx -- -pp "gospel pps")))
+```
+
+Be aware that it is the user's responsability to run the gospel type-checker when needed.
+
 ### Tools using Gospel
 
 > You are using Gospel as a frontend? [Let us
--- a/src/ppx_odoc_of_gospel/ppx_odoc_of_gospel.ml
+++ b/src/ppx_odoc_of_gospel/ppx_odoc_of_gospel.ml
@@ -1,0 +1,103 @@
+open Ppxlib
+
+let is_gospel attr = attr.attr_name.txt = "gospel"
+
+(* Prefix a string with whitespace the length of gospel special comment marker *)
+let align = "   "
+let eol_and_align = "\n   "
+
+let gospel_txt_of_attributes attr =
+  let rec aux attr =
+    if is_gospel attr then
+      match attr.attr_payload with
+      | PStr
+          [
+            {
+              pstr_desc =
+                Pstr_eval
+                  ( {
+                      pexp_desc = Pexp_constant (Pconst_string (txt, loc, _));
+                      _;
+                    },
+                    attrs );
+              _;
+            };
+          ] ->
+          { txt; loc } :: (List.map aux attrs |> List.flatten)
+      | _ -> []
+    else []
+  in
+  let rec last_loc_end = function
+    | [] -> failwith "unreachable case in last_loc_end"
+    | [ x ] -> x.loc.loc_end
+    | _ :: xs -> last_loc_end xs
+  in
+  match aux attr with
+  | [] -> None
+  | x :: _ as xs ->
+      let txt =
+        align
+        ^ String.concat eol_and_align ((List.map (fun { txt; _ } -> txt)) xs)
+      and loc_end = last_loc_end xs in
+      Some { txt; loc = { x.loc with loc_end } }
+
+let wrap_gospel header txt =
+  let header =
+    match header with
+    | `Declaration -> "Gospel declaration:\n"
+    | `Specification -> "Gospel specification:\n"
+  in
+  Fmt.str "{@gospel[\n%s%s]}" header txt
+
+let payload_of_string ~loc txt =
+  let open Ast_helper in
+  let content = Const.string ~loc txt in
+  let expression = Pexp_constant content |> Exp.mk ~loc in
+  let structure_item = Str.eval ~loc expression in
+  PStr [ structure_item ]
+
+let attr_label = function
+  | `Declaration -> "ocaml.text"
+  | `Specification -> "ocaml.doc"
+
+let doc_of_gospel header attr =
+  if is_gospel attr then
+    let attr_name = { txt = attr_label header; loc = attr.attr_loc }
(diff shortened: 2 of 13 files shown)


src/ppx_odoc_of_gospel/ppx_odoc_of_gospel.ml
```diff
@@ -0,0 +1,107 @@
...
+                    },
+                    attrs );
+              _;
+            };
+          ] ->
+          { txt; loc } :: (List.map aux attrs |> List.flatten)
+      | _ -> []
+    else []
+  in
+  let rec squash (acc : string loc) = function
```
shym, line 33: Loops of string concatenations always make me uneasy, `String.concat` is a lot more efficient as it doesn’t unnecessarily churn the GC.

shym (commented):
This looks pretty good, well done!

I have one main style suggestion: this is written first filtering Gospel attributes and then using `assert`s and `failwith`s in various places. Would it be really more cumbersome to drop all these assumptions and use `option`s instead, with a simple `filter_map` in the end, rather than a `filter |> map`? I think it would make further maintenance a lot more robust.
If you accept this suggestion, I would further propose:
- to add `ppx_odoc_of_gospel` in a single commit (in part in order to avoid a cumbersome splitting of the global modification of the module),
- but reuse all your nice commit messages as comments in the code, as they really explain its structure.

shym:
@n-osborne let me know if/when this is ready for a second review.

shym:
I tried running this code on `Gospelstdlib` but I lost the documentation (such as “The type for finite sequences.” for the type `'a sequence`):
- did I miss something?
- could we make it work?

shym:
A few ideas on how to make this work: as the OCaml compiler cannot attach those comments to an attribute, we would have to transform also those documentation comment, turning them into `gospel.text` attributes, for instance. As documentation comments are already lexed apart, that could be a minor change.

n-osborne:
Thank you for running the ppx against the gospelstdlib!

The problem is reproducible by removing an empty line between a documentation comment and a gospel declaration in `/test/ppx/odoc_of_gospel.mli`. Running the test will then show the documentation attribute disappear.

I believe your suggestion (@shym) should work (turning these documentation comments into `gospel.text` attributes in the source preprocessor and back again into documentation attributes in the ppx, IIUC). I'll work on implementing it.

That being said, we should be aware that if a user, for some reason, run the source preprocessor but not the ppx, the same documentation attributase will be lost from the documentation.

n-osborne:
I've created #324 to track this issue.

README.md
```diff
@@ -79,6 +79,20 @@ $ gospel check max_array.mli
...
+Gospel also provides a ppx rewriter to allow odoc to display the contents of
+gospel specifications and declarations as documentation. This ppx rewriter
+works under the assumption that the source preprocessor has been run first. The
+dune stanza read as follow:
+
+```dune
+(library
+ (name lib_name)
+ (preprocess
+ (pps gospel.ppx -- -pp "gospel pps")))
```
shym, line 91: Add an extra space for indentation.

README.md
```diff
@@ -79,6 +79,20 @@ $ gospel check max_array.mli
 OK
 ```
 
+Gospel also provides a ppx rewriter to allow odoc to display the contents of
+gospel specifications and declarations as documentation. This ppx rewriter
+works under the assumption that the source preprocessor has been run first. The
+dune stanza read as follow:
```
shym, line 85: → “reads”

src/ppx_odoc_of_gospel/ppx_odoc_of_gospel.ml
```diff
@@ -0,0 +1,109 @@
...
+        ( align ^ String.concat eol_and_align (List.rev (txt :: acc)),
+          loc.loc_end )
+    | { txt; _ } :: xs -> go (txt :: acc) xs
+  in
+  match aux attr with
+  | [] -> None
+  | [ x ] -> Some { x with txt = align ^ x.txt }
+  | x :: _ as xs ->
+      let txt, loc_end = go [] xs in
+      Some { txt; loc = { x.loc with loc_end } }
```
shym, line 45: The second case is covered by the third, isn’t it?

src/ppx_odoc_of_gospel/ppx_odoc_of_gospel.ml
```diff
@@ -0,0 +1,109 @@
...
+      | _ -> []
+    else []
+  in
+  let rec go acc = function
+    | [] -> failwith "unreachable case in go"
+    | [ { txt; loc } ] ->
+        ( align ^ String.concat eol_and_align (List.rev (txt :: acc)),
+          loc.loc_end )
+    | { txt; _ } :: xs -> go (txt :: acc) xs
+  in
```
shym, line 39: I would suggest replacing `go` with something along the lines of
```ocaml
align ^ String.concat eol_and_align (List.map (fun {txt;_} -> txt))
```
and
```
List.fold_left (fun _ { loc; _ } -> loc.loc_end) Lexing.dummy_pos
```
since `go` does two separate things at once.

shym (commented):
That’s nice to read, congratulations!
I’ve added a couple of small comments.

shym:
I suggest adding an empty interface for the ppx, to detect dead code.

test/ppx/dune
```diff
@@ -0,0 +1,21 @@
...
+ (libraries gospel.ppx ppxlib))
+
+(rule
+ (targets odoc_of_gospel.actual.mli)
+ (deps
+  (:pp pp.exe)
+  (:input odoc_of_gospel.mli)
+  (:gospel %{bin:gospel}))
+ (action
+  (run ./%{pp} --intf %{input} -pp "%{gospel} pps" -o %{targets})))
```
shym, line 16: `./` is not necessary in `./%{pp}`

shym (approved):
I made some small remarks on the way, but this looks good! Once those small details are ironed out, it’s ready to merge!

shym:
Oh! Now that I see the CI results, it might not be possible to use the `(pps ... -- -pp "...")` trick when the `pp` is built also in the project. At least I didn’t find a way to express the necessary dependency. We might have to do a few steps manually.

n-osborne:
Thanks for your comments.
Just pushed the modifications.

shym:
CI is happy, I merged.

PR #290, opened 2023-07-04
Support partial application of functions

shym:
This PR:
- adds support for partial applications of _functions_
- fixes syntax for applications of _constructors_
- adds a couple of examples, showing some of the changes it brings

Supersedes #284 as it does not change the AST.


src/typing.ml
```diff
@@ -317,11 +341,11 @@ let rec dterm kid crcm ns denv { term_desc; term_loc = loc } : dterm =
       let ls = find_q_ls ns q in
       if ls.ls_field then
         W.error ~loc (W.Symbol_not_found (string_list_of_qualid q));
-      if ls.ls_args <> [] then
-        W.error ~loc (W.Partial_application ls.ls_name.id_str);
-      let _, dty = specialize_ls ls in
-      let node, dty = (DTapp (ls, []), dty) in
-      mk_dterm ~loc node dty
+      if ls.ls_args <> [] then gen_app ~loc ls []
```
n-osborne, line 344: Maybe it's the opportunity to use a pattern matching here.

n-osborne (approved):
Great!
Thanks for the comments along the code :-)

Just had one tiny remark, feel free to ignore.

src/typing.ml
shym: Thank you for pinpointing that place, it made me realise that distinguishing cases is really not necessary there, `gen_app` will do exactly that special case anyway!

shym:
Fixed the issue you pointed and rebased.

PR #292, opened 2023-07-05
Improve generated `dune.inc` files

shym:
This PR:

- improves the dependencies declared for the test files: instead of
  using `(source_tree .)`, `gen.exe` now takes as arguments the
  dependencies (rarely used in the current test suite); it also
  explicit adds the gospel checker as a dependency, which should
  ensure all the tests are run again when gospel changes (it seemed to
  be an issue),
- adds a `@test-cmis` alias that will feed all the test `.mli` files
  to `ocamlc`, to ensure they are syntactically correct and, when
  relevant, well typed,
- generates cleaner `dune.inc`, by using a bit more of dune inference.

### DIFF
--- a/test/coercions/negative/dune.inc
+++ b/test/coercions/negative/dune.inc
@@ -1,55 +1,60 @@
 (rule
- (target bad_type_id.mli.output)
- (deps (source_tree .))
+ (deps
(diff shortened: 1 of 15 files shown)


n-osborne (approved):
Nice dune hacking.
Thanks!

PR #295, opened 2023-07-07
Display the backtrace of errors to help debug them

shym:


shym:
Updated to use a variable `GOSPELDEBUG` so that setting `OCAMLRUNPARAM=b` does not automatically trigger that behaviour in Gospel.

PR #297, opened 2023-07-10
Support patterns of one-parameter constructors with a tuple argument

### LINKED ISSUE #210: Spurious type check error
The `gospel check` command fails on input :
```
val f: int -> (int * int, int) result
(*@ r = f ()
    ensures
      match r with
      | Ok (p, q) -> true
      | _         -> false *)
```

with the following spurious error :
```
Error:
  The constructor `Ok' expects `1' argument(s)
  but is applied to 2 argument(s) here.
```

It may be related to too late instanciation of type variables from `('a*'b) result`.

shym:
This PR brings support for patterns of constructors expecting one argument that can be a tuple such as `Some (x,y)`.

Closes #210


n-osborne (approved):
Thank you!
LGTM

n-osborne:
About tests' location, I would suggest to update them in a different commit for now, postponing the decision on what we do with the positive/negative distinction when dealing with #298

shym:
Just rebased on `main` to fix the formatting issues that were from `main` itself.

PR #299, opened 2023-07-11
Fix specifications location

n-osborne:
Takes the location of the payload instead of the attribute
so that the length of `"(*@"` is not lost along the way.


shym:
Very nice!
This can be rebased on `main` to scrap `Tast` reformatting from that PR. And then that can be merged in!

shym (approved):

PR #300, opened 2023-07-11
Check that all patterns in a disjunction bind the same variables

### LINKED ISSUE #224: Missing error : variable in sub-pattern
I think the following program should raise an error since `a` variable isn't bind in both sides of the or-pattern.

```ocaml
val f : int -> int
(*@ y = f x
      ensures match x with _ | a -> a = 1 *)
```

shym:
During type-checking, ensure that all branches of a disjunctive pattern bind the same set of variables.

Closes #224

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,6 +57,8 @@
 
 ## Fixed
 
+- Check that all patterns in a disjunction bind the same variables
+  [\#300](https://github/ocaml-gospel/gospel/pull/300)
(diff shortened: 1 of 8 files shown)


n-osborne (approved):
Thank you.
Need formatting before merging, but otherwise very good.

shym:
Rebased and reformatted.

shym:
Rebased

shym:
Rebased

Issue #296, opened 2023-07-07
Allow tuple destruction in anonymous function


n-osborne:

shym:
A reasonable implementation would probably use patterns there, as we would want to allow also nested tuples, etc.
But patterns do not parse casts at the moment it seems (and they would be required for function arguments):
```ocaml
val f : int option -> int
(*@ r = f x
    requires match x with
             | Some (y:int) -> y >= 0
             | None -> true *)
```

PR #301, opened 2023-07-12
Fix pattern with cast

n-osborne:
As mentioned by @shym [here](https://github.com/ocaml-gospel/gospel/issues/296#issuecomment-1629102711).

This PR proposes to simply drop the type information as the translation from destructive patterns to typed patterns is done after typing.

The other possibility would be to add a `Pcast` to `Tterm.pattern_node`, but we expect the type information of the cast to have been put in the casted pattern.


shym (approved):
LGTM, very well done!
I just wondered a bit about the commit message of a8382cf, I don’t understand the last part of the sentence.

PR #303, opened 2023-07-13
Add two new issues

shym:


shym:
For the record, the error found in CI is due to the rendering of `assert` failures that changed between OCaml 4.11 and 5.0.
This happens in issue `t2.mli`, so that error is found in all the tests since it was merged.

PR #304, opened 2023-07-13
Take recursivity into account when typing type declarations

### LINKED ISSUE #225: Type declaration for contains a cycle error
`gospel check` issues a `The type declaration for `t' contains a cycle` error for the following piece of code:
```
module type A = sig
  type t
end

module type B = sig
  type t

  module C : A with type t = t
end
```

I believe this is a bug in the type-checker.

shym:
Before, type declarations were considered always recursive.
Route the recursivity information to the function typing type declarations and feed it `Nonrecursive` when the function is reused to type module constraints

Add the (now positive) example that was reported in #225.

Add, while we’re at it, what should be done to fix issue t6.

Closes #225

### DIFF
--- a/src/typing.ml
+++ b/src/typing.ml
@@ -516,7 +516,7 @@ let process_type_spec kid crcm ns ty spec =
     spec.ty_loc
 
 (* TODO compare manifest with td_kind *)
-let type_type_declaration kid crcm ns tdl =
+let type_type_declaration kid crcm ns r tdl =
   let add_new tdm td =
     if Mstr.mem td.tname.txt tdm then
       W.error ~loc:td.tname.loc (W.Name_clash td.tname.txt)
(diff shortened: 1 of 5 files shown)


src/typing.ml
```diff
@@ -684,7 +687,7 @@ let type_type_declaration kid crcm ns tdl =
   tdl
 
 let process_sig_type ~loc ?(ghost = Nonghost) kid crcm ns r tdl =
-  let tdl = type_type_declaration kid crcm ns tdl in
+  let tdl = type_type_declaration kid crcm ns r tdl in
```
n-osborne, line 690: Why don't you use `Tast.rec_flag` here (with the `rec_flag` converter used the line below)?

n-osborne (approved):
Just one tiny remark. I think it would be better to use only one rec_flag type.

PR #307, opened 2023-07-19
Add some (failing) tests about anonymous functions

n-osborne:
Anonymous functions does not accept yet patterns.

Co-authored-by: Samuel Hym <samuel.hym@rustyne.lautre.net>

### DIFF
--- a/test/patterns/dune
+++ b/test/patterns/dune
@@ -1,0 +1,15 @@
+(include dune.inc)
+
+(rule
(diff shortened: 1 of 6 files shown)


shym:
I’ve just added a couple of syntax fixes in the tests, thanks to a branch which supports patterns, so that Gospel can report other syntax errors instead :-)

shym:
I’ve slightly altered the commit message before merging :-)

PR #308, opened 2023-07-20
Propagate pattern locations to the final representation

shym:
Preserve locations in patterns during the last translation (from `dpattern` to `Tterm.pattern`)
Allow the typechecker to pinpoint the precise locations of errors, rather than pointing to the whole enclosing expression
Update the tests accordingly

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,8 @@
 
 ## Improved
 
+- Propagate pattern locations to report errors to the precise patterns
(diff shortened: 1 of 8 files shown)


n-osborne:
Nice, thank you!

CI failure is caused by experimental features or turned off switch, so it is good to merge.

As a side note, while reviewing, I've noticed that the code base is not very consistent regarding the position of the location arguments (mostly the last one, but sometimes not, and sometimes named argument...)

n-osborne (approved):

Issue #305, opened 2023-07-18
Gospel does not parse `fun` type annotations as OCaml


shym:
The following code is accepted by the OCaml type checker:

```ocaml
  fun true : int -> 12
```

(besides the obviously non-exhaustive pattern `true`), meaning that the `int` is the return type but the Gospel typechecker accepts the following:

```ocaml
val f : int list -> int
(*@ y = f xs
    ensures List.for_all (fun x : int -> true) xs *)
```

where `int` is interpreted as the type for the argument `x`.

shym:
#309 will change that situation:
```
val f : int list -> int
(*@ y = f xs
    ensures List.for_all (fun x : int -> true) xs *)
```
will now report a syntax error at `: int` as the type annotation is not handled by the parser.

n-osborne:
Fixed by #309

PR #309, opened 2023-07-20
Use patterns for anonymous-function arguments

### LINKED ISSUE #296: Allow tuple destruction in anonymous function

shym:
Decouple `fun` from quantifiers (`forall` and `exists`) so that it reuses the usual OCaml syntax, namely accepting patterns rather than just variable names
As in `match ... with ...`, reject non-exhaustive patterns

This makes it possible to destruct tuples by `(fun (x,y) -> ...)`

This PR is based on #308 to get proper locations for the errors.
I’ll add the changelog entry when the PR is closer to merging, to avoid the usual conflicts.

Closes #296


shym:
I’ve updated this PR with a modification of the grammar to handle the example shown in #305 as OCaml would. So it would close #305 also.
That example should be added to this PR, but it’d better wait for the upcoming reorganization of the tests to go through.

n-osborne:
Nice addition, thank you :smiley: 

LGTM

n-osborne (approved):
Nice addition, thanks :smiley: 

LGTM

PR #311, opened 2023-07-21
Tell git to consider `dune.inc` files as binaries

shym:
Tell git to stop considering generated `dune.inc` files as source files that can be diffed so that it will:

- skip that noise in diffs, logs, etc.
- leave the content of those files untouched when there is a merge conflict, as conflict markers prevent `dune` from interpreting those files, for instance to run `dune runtest --auto-promote` that would fix them...


shym:
This has been included in the bigger #312 PR, so #311 served its testing and reviewing purpose.

PR #313, opened 2023-08-12
Update README for opam release

sim642:
This just updates README based on the same text in documentation: https://ocaml-gospel.github.io/gospel/getting-started/installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,10 +44,9 @@ Gospel!
 
 ### Installation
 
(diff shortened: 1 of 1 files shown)


shym:
Side-note: Gospel (the language and the tool) changed quite a bit since the version released in opam.
A new 0.2 release should be coming shortly :crossed_fingers:

PR #312, opened 2023-07-25
Test cleanup

n-osborne:
- Share as much as possible test/*/dune logic
- Move the test dune-config generator to `test/utils/dune_gen.ml`
- Document the commandline arguments for dune_gen
- Move the test checker to `test/utils/testchecker.ml`
- Add the dependency to the gospel binary for every test
- Add an alias `duneincs` to generate the `dune.inc` files
- Tell git to consider `dune.inc` files as binaries
- Document how to have git call dune to resolve conflicts on `dune.inc` files
- Cleanup tests part one

Issue #192, opened 2022-05-10
disallow ambiguous or-patterns under guards


backtracking:
On such a pattern the OCaml compiler emits warning 57:
```
let f = function
  | (x, 0) | (1, x) when x = 0 -> 3
  | _ -> 4
```
that is,
```
Warning 57: Ambiguous or-pattern variables under guard;
variable x may match different arguments. (See manual section 9.5)
```
See https://v2.ocaml.org/releases/4.14/htmlman/patterns.html#sss:pat-or

Suggestion: do not allow such patterns in Gospel.

mariojppereira:
Are you suggesting to completely disallow such patterns, or only if there is a `when` clause ?

What is the current status of `when` inside Gospel patterns ?

backtracking:
Only when there is a `when` clause combined with a or-pattern.

backtracking:
Regarding `when`, @paulpatault is currently adding exhaustiveness checks for patterns in Gospel, and this will include `when` guards as well.

mariojppereira:
Great to hear `when` clauses are coming to Gospel :D

mariojppereira:
> On such a pattern the OCaml compiler emits warning 57:
> 
> ```
> let f = function
>   | (x, 0) | (1, x) when x = 0 -> 3
>   | _ -> 4
> ```
> 
> that is,
> 
> ```
> Warning 57: Ambiguous or-pattern variables under guard;
> variable x may match different arguments. (See manual section 9.5)
> ```
> 
> See https://v2.ocaml.org/releases/4.14/htmlman/patterns.html#sss:pat-or
> 
> Suggestion: do not allow such patterns in Gospel.

mariojppereira:
> Only when there is a `when` clause combined with a or-pattern.

I agree with your initial suggestion to simply disallow such patterns in Gospel.

PR #316, opened 2023-08-25
Rephrase ambiguous-patterns error

### LINKED ISSUE #314: Remove “ambiguous patterns” detection?
A code such as:

```ocaml
match x with
| _ | _ when x = 1 -> true
| _ -> false
```

is rejected by Gospel as it judges it as “ambiguous” due to the alternative under a guard (`_ | _ when ...`). This is not ambiguous as far as OCaml is concerned: on that same example, OCaml will just warn about the redundancy.

I propose to drop that detection to keep closer to the OCaml behaviour if there is no ulterior motive for that error.

shym:
Alternative patterns under guards are not considered ambiguous by the OCaml compiler so accept them also in Gospel patterns

This is intentionally based on #312, to see the impact on the cleaned-up test suite. Only one test ran into this feature.

Closes #314

### DIFF
--- a/src/warnings.ml
+++ b/src/warnings.ml
@@ -138,7 +138,10 @@ let pp_kind ppf = function
         "This pattern-matching may not be exhaustive because of the guard.@\n\
          Here is an example of a case that may not be matched:@\n\
         \  %s" p
-  | Ambiguous_pattern -> pf ppf "Ambiguous or-pattern under guard"
+  | Ambiguous_pattern ->
+      pf ppf
+        "Or-patterns are prohibited under guards to avoid ambiguities@ (see \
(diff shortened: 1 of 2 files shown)


shym:
Following the discussion in #192, I rewrote this PR to just rephrase the error and to point to the OCaml compiler warning. Raising that error in exactly the same cases as OCaml would require detecting whether a given variable is placed always at the same position in all alternatives of the “or” pattern. So:
- to do this properly would require more work,
- combining or-patterns and guards is arguably somewhat ambiguous in the first place.
A better error message to guide users is a good compromise I think.

shym:
Rebased and up for review.

n-osborne (approved):

PR #319, opened 2023-08-28
Fix issues with source location tracking (directives and overridden filename)

shym:
- Take directives into account while preprocessing
  Update the current position of the lexer if a `# line "file"` directive is found while preprocessing, to follow the behaviour of the OCaml lexer
- Preserve the filename found during lexing while parsing Gospel
  The lexer is careful to report locations, override the filename (with the filename fed on the command-line) only if the lexer position is incomplete

This PR does not support (yet?) directives inside Gospel code, as I’m not sure about the interactions with OCaml, when line directives would be stored inside an attribute: I guess the Gospel and OCaml lexers would easily get out of sync again.

This seems to solve the problems mentioned in #253.


n-osborne (approved):
LGTM
Thank you.
I believe this deserves an entry in the changelog.

n-osborne:
Apparently, it needs formatting too.

shym:
As CI was finally happy, I merged.

PR #321, opened 2023-08-29
Consider comments as spaces while preprocessing

### LINKED ISSUE #320: Comments can break preprocessing
```ocaml
(*@ type t *)
(*@ ephemeral *)
```

is correctly preprocessed into:

```ocaml
[@@@gospel
 {| type t |}[@@gospel
 {| ephemeral |}]
               ]
```

(note the nesting) but

```ocaml
(*@ type t *) (* some type *)
(*@ ephemeral *)
```

is preprocessed into:

```ocaml
[@@@gospel {| type t |} ] (* some type *)
[@@gospel {| ephemeral |} ]
```

which is not properly nested (and not correct OCaml syntax).

This is breaking quite a few examples from the documentation.

shym:
In the following specification:

```ocaml
(*@ type t *)
(*@ ephemeral *)
```

the preprocessor needs to recognize that `ephemeral` must be attached in a nested attribute to the first attribute containing `type t`. It does so by accepting `Space` between the two items, but not simple comments.

So this turn simple comments into another form of `Space` so that

```ocaml
(*@ type t *) (* a ghost type *)
(*@ ephemeral *)
```

is valid syntax.

Closes #320 (hopefully)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -61,6 +61,9 @@
 
 ## Fixed
 
+- Consider comments as spaces while preprocessing (to ensure specification can
+  be attached to a ghost function or type, for instance)
+  [\#321](https://github/ocaml-gospel/gospel/pull/321)
 - Fix source-location tracking (directives and overridden filename)
   [\#319](https://github/ocaml-gospel/gospel/pull/319)
 - Set up location in parsing ghost specifications
--- a/src/pps.mll
+++ b/src/pps.mll
@@ -193,9 +193,12 @@ rule scan = parse
     }
   | "(*"
       {
+        push ();
         Buffer.add_string buf "(*";
         comment lexbuf;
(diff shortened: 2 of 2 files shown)


n-osborne (approved):
Nice catch!

As a sidenote: it does not work with documentation attributes, which is to be expected but maybe we should include this information in the documentation.

shym:
Where would you expect this to be documented?
Maybe this would be fixed as part of a #324 fix?

n-osborne:
#331 add support for

```
(*@ type t *)
(** documentation *)
(@ ephemeral *)
```
with any number of newline character between the ghost declaration and its specification.

n-osborne:
This just need an entry in the changelog and it is good to be merged.

shym:
I added a CHANGES entry and CI is happy, so I merged.

Issue #57, opened 2021-10-07
Publish a full documentation


pascutto:
Still to do:
- [x] Finish the mutable queue walk-through
- [x] Finish the union-find walk-through
- [ ] Write a landing page
- [ ] Polish the language specification section
- [x] Finish the FAQ

jmid:
*logical.md*

- 'Uninterpreted symbols and Axioms' is incomplete.
   There is an axiom example in 'Logical function contracts' however, so perhaps a first version can be completed by rearranging it.

- I was missing an explanation of `variant`, e.g.,

	"A `variant` is a termination measure. It has to have type `int`, be non-negative, and decrease strictly on each recursive call. "

  (but this is just me guessing...)

PR #323, opened 2023-09-01
Update documentation

### LINKED ISSUE #293: Check that the documentation is up to date

### LINKED ISSUE #322: Move the documentation about the attribute syntax to an appendix page
The section “Language specification” of the documentation starts with the explanation of the attribute syntax, etc. which is never used, and is ever used only by Gospel developers, probably.
I suggest moving it to an appendix page.

shym:
This PR updates the documentation. It is based on #253, as it changes the fixed version of the documentation, obviously.

It does a large reorganisation of the language reference part.
It addresses the comments in #57 that were not yet taken into account.

Closes #322

It should contribute to close #293


shym:
Latest updates to that PR bring among other things a build of the documentation as an CI artifact, so that the actual rendering can be checked locally (it requires a web server, as docusaurus URLs are not real paths) before merging to `main` and actually deploying it.

shym:
Latest update: I added the suggested link from the Welcome to Stdlib, and dropped the Tips page, after moving the tips it contained to the relevant sections.

PR #325, opened 2023-09-06
Reformulate "Symbol not found" error to mention scope

### LINKED ISSUE #318: `Symbol not found` is sometimes misleading
The message `Symbol not found` is used in many cases, some of which I find rather hostile to newcomers.

On the following example:

```ocaml
(*@ val g: int -> int *)

val f : int -> int
(*@ r = f x
    ensures r = g x *)
```

`gospel check` reports:

```
File "abc.mli", line 5, characters 16-17:
5 |     ensures r = g x *)
                    ^
Error: Symbol g not found.
```

This is due to the fact that `g` was not declared as `pure`.

The same message is used for `r` in the following:

```ocaml
val f : int -> int
(*@ r = f x
    requires r = 0 *)
```

where the result is bound in a `requires`.

Maybe specific messages should be used in those cases to help users understand what is in scope.

shym:
Point users to the documentation in which scopes are used for which clauses

Requires #323 to bring the said documentation page

Closes #318

### DIFF
--- a/src/warnings.ml
+++ b/src/warnings.ml
@@ -109,7 +109,11 @@ let pp_kind ppf = function
   | Invalid_int_literal (s, c) ->
       pf ppf "Invalid int literal: `%s%a'" s (option char) c
   | Symbol_not_found sl ->
-      pf ppf "Symbol %a not found" (list ~sep:(const string ".") string) sl
+      pf ppf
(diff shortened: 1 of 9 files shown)


n-osborne (approved):
As discussed offline, this is the best solution without having code clever enough to figure out why the symbol is out of scope.

Just need some formatting and it is good to go.

shym:
CI is happy, I merge.

PR #326, opened 2023-09-07
Introduce a generic `pp_gen` pretty-printer for error messages

shym:
Generalise the error pretty-printer so that its logic to fix the locations (due to preprocessing messing them up) can be reused directly by other tools using gospel as a library

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 ## Added
 
+- Introduce a generic `pp_gen` pretty-printer for error messages, so that
+  external tools can pretty-print errors in the same style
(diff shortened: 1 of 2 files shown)


shym:
I rebased and changed the name of the new generic function, as `pp_fix` was too bound to the fact that it is fixing locations, while it also opens the way to displaying warnings.

n-osborne:
This deserve an entry in the changelog, but I agree that `pp_gen` is a better name than `pp_fix`.

shym:
I added a CHANGES entry and, as CI was happy, merged it.

PR #327, opened 2023-09-08
Add a dune alias to check that `dune.inc`s files mention only git-tracked files

shym:
Add a `pre-push` dune alias to check that all the files appearing in `dune.inc`s are indeed tracked in `git`, to (help) avoid pushing `dune.inc`s generated for local-only files.


shym:
This is brittle and incomplete (see #328 for something it would have missed).

PR #328, opened 2023-09-11
Remove outaded dune rule

n-osborne:
Dune rule for removed `good-practice.md` page was still there.


shym:
Sorry for that! :see_no_evil:
And thank you for the fix.

PR #329, opened 2023-09-11
Add a Github Action CI workflow to run all the tests

shym:
This PR adds a new CI workflow that runs all the tests in an action.

The goals are to:
- test the installation of the OPAM package (not tested by OCaml-CI) to make sure we are ready before we release,
- test the noisy tests (that are not attached to `@runtest` because they are noisy),
- provide a fallback for when OCaml-CI is overwhelmed.

This should provide a more reliable way to avoid broken states than what #327 tried to introduce.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,41 @@
+name: Build and test
+
+on: [push, pull_request]
+
+jobs:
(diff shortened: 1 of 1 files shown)


n-osborne (approved):
Good idea !

PR #331, opened 2023-09-12
fix pps documentation of ghost

### LINKED ISSUE #324: gospel pps still detaches some documentation attributes
```
$ cat > foo.mli << EOF
> (** documentation *)
> (*@ function f : (i : integer) : integer *)
> EOF
$ ocamlc -dsource -pp "gospel pps"
[@@@gospel {| function f : (i : integer) : integer |}]
```

The documentation comment should be turned into a documentation attribute by `gospel pps`.

n-osborne:
This PR add support for the documentation of ghost declarations.
Gospel preprocessor should now accept every interleaving of ghost declaration, with and without specification and with or without spaces between these items. Spaces are checked to contains not more than one end of the line character when necessary.
Decision has been made (but can obviously been discussed) to support documentation between a ghost declaration and its specification with no constraints on the space in between. (See tests added in the first commit for examples)

Fixes #324


shym (commented):
I’ve given this PR quite some thought: it is consistent with the current state of the code base but I find it troubling that we end up with so much duplicated code in `print`.
Without going full steam ahead with a complete rewrite of the preprocessor (as required to address #336, but that’s for another PR entirely :sweat_smile:), I’d suggest the following trick (also consistent with the current code base, I’d say): we could tweak `collapse_spaces` to ensure there is always exactly one (possibly empty) `Spaces` in between any other tokens. If I’m not mistaken, simply dropping lines 207 and 211 would do the trick. That would allow de-duplicating a lot of cases in `print`, and help making sure we are not forgetting some cases.

I also wondered about the choice of filtering on `Spaces _` between `Ghost` and `Spec`: my first intuition would be to mimick OCaml’s behaviour with documentation, i.e. attaching a `Spec` to its predecessor only when there is no empty line in between.

By the way, `is_small` is not exactly mimicking OCaml because of a tricky case (following #321). In the following extract:

```ocaml
val f : int -> int
(*

*)
(** bla *)
```

`bla` is correctly attached to `f`, even if there are various `'\n'` between them.

On a completely different level, I would suggest gently asking `vim` to `:2,220!ocamlformat --impl - | sed 's/^./  &/'` to format the OCaml code within `pps.mll`, for legibility (and consistency with the rest of the project).

shym:
I updated this PR with one large modification. Now determining whether space should separate two consecutive items (aka be considered `Large`) is using data from the lexer (which rules were applied) to handle possibly-nested comments and contained strings properly.

On the way, two minor updates:
- a rebase on `main`,
- reindenting (two spaces, for consistency).

shym:
I just force-pushed an update that only fixes a few typos in a commit message.
I’ll merge if CI is happy.

PR #333, opened 2023-09-14
Use standard lowercase `true` and `false` for booleans

### LINKED ISSUE #315: Why `fs_bool_false` is `False` rather than `false`?
`fs_bool_false` and `fs_bool_true` symbols use capitalized spelling `False` and `True` rather than the standard OCaml booleans. It makes error messages confusing.

Is that another symptom of the confusion between `prop` and `bool` (see #287)?

I propose to use standard spellings.

shym:
Parsing used the standard lowercase syntax but pretty-printing was using uppercase versions

Closes #315

### DIFF
--- a/src/symbols.ml
+++ b/src/symbols.ml
@@ -84,12 +84,12 @@ let fs_unit =
 
 let fs_bool_true =
   fsymbol ~constr:true ~field:false
-    (Ident.create ~loc:Location.none "True")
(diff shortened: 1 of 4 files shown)


n-osborne (approved):
Just need an entry in the changelog and can be merged.

shym:
This seemed more a typo than an explicit choice, so I think we can skip the changelog.

Issue #302, opened 2023-07-13
`Tmodule.add_sig_contents` silently ignores unsupported signature items


shym:
At the moment `Tmodule.add_sig_contents` ends up with a catch-all case that silently ignores unsupported signature items (such as what happens with `include`). An error message or warning might be more helpful, until all cases are indeed supported?

PR #334, opened 2023-09-14
Display a warning when encountering an `include`

shym:
This is based on #326 to have access to the `pp_gen` pretty-printer.

That would be a (partial) way to address #302.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,8 @@
 
 ## Improved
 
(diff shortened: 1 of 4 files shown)


n-osborne (approved):
Just need an entry in the changelog and it is good to be merged.
Thanks :smiley:

shym:
I added the CHANGES entry and CI is happy, so I merge it.

PR #339, opened 2023-09-21
Improve type-error messages for mismatches between header and val type

shym:
Distinguish between parameters and return values
Handle the special case of too many parameters
Fix a typo in the mismatch error message


n-osborne (commented):
Thanks, these are indeed better error messages.

I'm a bit doubtful about the returned value case. As a programer, I would expect only one returned value. So a message stating that there are to few of those sound a bit strange. But we also have this error message when a returned tuple is not destructed:

```
val f : int -> (bool * int)
(*@ b = f i *)
```

will raise:
```
File "/tmp/bob.mli", line 2, characters 8-9:
2 | (*@ b = f i *)
            ^
Error: Type checking error: too few returned values.

```

shym:
I confess myself lazy there...
I rewrote the function to test, in that particular case, whether we are testing the returned values for a function returning a tuple and, if so, expand the explanation. It might now err on the side of verbosity.
I’ve added a test for that particular case.

PR #341, opened 2023-09-21
Use a style for error messages closer to the OCaml compiler

shym:
- do not use quotes for code extracts,
- use magenta for warnings.


n-osborne (approved):
Nice, thank you. This just need some reformating and it is good to go.

shym:
Reformatted and merged.

Issue #337, opened 2023-09-19
OCaml compiler reports error on code generated by the preprocessor when the attribute is not at an acceptable location


shym:
Given the following source file `middle.mli` with a Gospel specification appearing at an improper location (in the middle of a type):

```
# 1 "middle.mli"

val f : int ->
(*@ x = f y *)
int
```

the user will see an error reported against the generated part of the file:

```
$ ocamlc.opt -pp "_build/install/default/bin/gospel pps" middle.mli
File "middle.mli", line 3, characters 0-3:
3 | [@@gospel
    ^^^
Error: Syntax error
```

It would be nicer to be able to detect such cases and turn them into a `[@@@ocaml.warning` similar to OCaml warning 50.

n-osborne:
Maybe we should use the `[@@@ocaml.ppwarning]` attribute. It is documented as the one to use in preprocessor.

`gospel pps` should then turn the `middle.mli` file into something like:

```ocaml
# 1 "middle.mli"

[@@@ocaml.ppwarning "Gospel specification in the wrong place"]
# 3 "middle.mli"
val f : int ->
(*@ y = f x *)
int
```

And we will have:

```sh
$ ocamlc.opt middle.mli

**Comment by n-osborne** [comment]
Maybe we should use the `[@@@ocaml.ppwarning]` attribute. It is documented as the one to use in preprocessor.

`gospel pps` should then turn the `middle.mli` file into something like:

```ocaml
# 1 "middle.mli"

[@@@ocaml.ppwarning "Gospel specification in the wrong place"]
# 3 "middle.mli"
val f : int ->
(*@ y = f x *)
int
```

And we will have:

```sh
$ ocamlc.opt middle.mli
File "middle.mli", line 2, characters 20-61:
2 | [@@@ocaml.ppwarning "Gospel specification in the wrong place"]
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning 22 [preprocessor]: Gospel specification in the wrong place

```

shym:
Indeed, well spotted!

shym:
Sidenote: unfortunately, `ppwarning` will trigger a warning during OCaml _typing_, not _parsing_. And gospel relies only on parsing. So if we generate such a warning, it won’t appear in a `gospel check`. :thinking:

PR #348, opened 2023-09-25
Add a test that shows the issue of misplaced Gospel comments

shym:
As Gospel special comments are turned into attributes, if they are placed at positions at which the OCaml syntax does not accept attributes, they will be reported as syntax errors, which might be confusing to new users.

It might be useful to mention this fact also in the documentation page explaining the Gospel special comments.


shym:
This is to keep track of the issue reported in #337 in the test suite.

n-osborne (approved):

Issue #336, opened 2023-09-18
Preprocessor uses a _lot_ of memory (and consequently of GC time)


shym:
To test the performance of `gospel pps`, I’ve generated `.mli` files by concatenating copies of `test/vocal/Vector.mli` in a format:

```ocaml
module Block0 = sig
(* copy of Vector.mli *)
end
module Block1 = sig
...
```

Results:

- 16 copies (~235KB `.mli` file):
  ```
  allocated_words: 439664837
  real       0m1.798s
  ```
  with 439664837 = 27479052 × 16

- 32 copies (~471KB `.mli` file):
  ```
  allocated_words: 1748175205
  real       0m11.291s
  ```
  with 1748175205 = 54630475 × 32

- 48 copies (~700KB `.mli` file):
  ```
  allocated_words: 3766420005
  real       0m23.129s
  ```
  with 3766420005 = 78467083 × 48

(So note that the allocated_words and time are not linear)

shym:
#353 might be enough to fix this, even though it still loads the whole source into memory. Otherwise we should go into a fully streaming preprocessor.

Issue #342, opened 2023-09-22
Add a directive in the beginning of the preprocessed file


n-osborne:
When OCaml calls the preprocessor it expects a directive in order to recover locations.

PR #353, opened 2023-10-03
Improve performance and clean up the preprocessor

shym:
This PR reworks the preprocessor by:

- cleaning up with a more complete and uniform reindentation,
- improving a lot its performance by generating the output into a formatter rather than generating lots of temporary strings.

In a bit more details, the performance gains are obtained by changing the interface of the preprocessor so that it outputs its result into a formatter rather than a string and propagating the change all the way down, so that:
- the whole output is generated in a single buffer when the target is a `string`, or directly on `stdout` in the case of `gospel pps`,
- `print` becomes tail recursive.

Using the example mentioned in #336 (test run in CI, so the performance may vary depending on load...), we go from:

```
16
allocated_words: 440218559
real	0m1.814s

32
allocated_words: 1749282355
real	0m10.295s

48
allocated_words: 3768016163
real	0m16.260s
```

to:

```
16
allocated_words: 2152756
real	0m0.019s

32
allocated_words: 4253492
real	0m0.031s

48
allocated_words: 6354228
real	0m0.044s
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -68,6 +68,8 @@
 
 ## Fixed
 
+- Fix the performance issues in the preprocessor
+  [\353](https://github/ocaml-gospel/gospel/pull/353)
 - Gospel preprocessor support documentation for ghost declaration
   [\#331](https://github/ocaml-gospel/gospel/pull/331)
 - Consider comments as spaces while preprocessing (to ensure specification can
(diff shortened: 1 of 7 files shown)


shym:
Testing on larger inputs seem to show about 14MB/s so this is probably good enough for now to say:
Closes #336.

shym:
Closes #342

shym:
Updated so that all `print_X` functions (not `print` itself, though, which is using) take `ppf` as their last argument. And rebased.

n-osborne (approved):
Great, thank you!

PR #356, opened 2023-10-05
Add dune rules to generate documentation for the Gospelstdlib and fix its syntax

shym:


n-osborne (approved):
Great! Thank you!

n-osborne:
Need formatting.
And apparently, ocaml-ci doesn't know about `man` anymore...

PR #357, opened 2023-10-06
Add release number in CHANGES

n-osborne:


shym (approved):
It can be released, indeed!

PR #359, opened 2023-10-06
Updates after the release

shym:
This PR updates:
- the README with the newly-released version number
- the documentation with the rendering of the `Gospelstdlib` (by the gospel.ppx+odoc+odig pipeline)

and add a link to that local rendering of the Gospelstdlib. As docusaurus is still utter magic to me, I’m not sure the link will actually work when deployed, though :-/


n-osborne (approved):
Nice. Thank you.
I have no idea about the link neither...

PR #362, opened 2023-10-30
Run tests on the new example

shym:


n-osborne (approved):
Thanks, good idea!

I wonder whether the `gospel_expected` message added at the end of the interface could be misleading to people reading the examples as learning ressources.
Maybe a silenced `gospel check` could be enough?

mariojppereira:
Hi @shym and @n-osborne, sorry for my (very) late answer. Actually, I was thinking about the folder `examples` to be a sort  of repository for "ongoing" case studies. This means you can use put there gospel specification that *does-not-even-yet* type-checks. So, I guess always running sanity tests on these examples might not be the best idea.

Maybe I should try to hide this folder, or maybe give it a more suggestive name?

n-osborne:
comment.md

n-osborne:
Thanks @mariojppereira.

Indeed, as a newcomer, a folder named `examples` would be the first place I look to learn how to write *correct* specifications. And I really think it is a very good idea to have that :-)

For specifications that are not (yet) corrrect, the `test/issues` folder may be a good place? We added it with @shym some time ago to track tests related to github issues (or at least, that was the idea, not completely sure we have been thorough with that).

Another idea would be an `experiment` folder (maybe `test/experiment`?).

shym:
I agree a different name would be better for WIP specs. I don’t mind that directory being visible as long as its purpose is clear (by its name and by a README).
Even for WIP specs where the typechecker will (initially) fail, I think it is important to keep track of it. Note that the tests we run do not _require_ the file to type-check, but will just add a prologue with the `gospel check` error if any. This will be useful to help develop the typechecker to cover those new cases.

n-osborne:
Long overdue. Merging.

PR #363, opened 2023-11-06
Remove the `gospel_expected` prologue at the end of successful tests

shym:
Wrapping up the `OK` of `gospel check` up in a `gospel_expected` comment is verbose and not really useful, so drop them altogether in the test suite
Address a comment made in #362


n-osborne (approved):
Good idea.
I'm just wondering whether the `is_ok` test is really necessary as `status`  shouldn't be `0` when type checking fails and output should be on stderr.

shym:
I did this just in case, but you’re right, if status is 0, the output is always OK.
So I rewrote the code to directly ignore stdout.

PR #368, opened 2023-11-27
Update ocamlformat to 0.26.1

n-osborne:


shym (approved):
As CI agrees this has no impact on the formatting of the current code base, let’s merge it.

PR #372, opened 2023-12-12
Make `with` necessary when declaring type invariants.

### LINKED ISSUE #369: `with` is optional for type invariants
Gospel parser allows for:
```ocaml
type t
(*@ invariant ... *)
```
and
```ocaml
type t
(*@ with x invariant ... *)
```
Is the first form necessary?
My feeling is that without naming a value of type `t`, there is not much content to give to a type invariant. But maybe I am missing something?

This then make the identifier optional in the AST, hence an additional layer of destruction (so to speak) when exploring a type specification and a decision to make about what to do when there is no name (which is useless work if this never happen, and hard to do without having an idea of a potential use case).

Also, I believe the documentation is not up to date: https://github.com/ocaml-gospel/gospel/blob/main/docs/docs/language/type-specifications.md?plain=1#L33

n-osborne:
Fixes #369

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,11 @@
   to experiment with Gospel syntax and future extensions of the language.
   [#361] (https://github.com/ocaml-gospel/gospel/pull/361)
 
+## Improved
+
+- Make the `with` necessary when declaring type invariants
+  [\#372](https://github.com/ocaml-gospel/gospel/pull/372)
+
 ## Internals
 
 - Remove the `gospel_expected` prologue at the end of successful tests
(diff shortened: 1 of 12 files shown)


shym (approved):
LGTM, well done!
It requires a round of `dune runtest --auto-promote` or two before merging.
A small suggestion: it might be nicer in `src/upretty_printer.ml` to reset `Option` with a `module Option = Stdlib.Option` just after the `open`s, since we don’t really need `Opprintast.Option` anyway.

n-osborne:
Thanks.
I've followed your suggestion and updated the documentation to reflect the change.

shym (approved):
I pushed 2 mini-fixes. They should be squashed and this is good to merge for me!

n-osborne:
CI failure is unrelated.
I merge.

PR #374, opened 2024-01-08
Propagate non-optional name to typed ast

n-osborne:
This PR completes #372.

It contains a breaking change in the typed AST.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,8 @@
 ## Improved
 
 - Make the `with` necessary when declaring type invariants
-  [\#372](https://github.com/ocaml-gospel/gospel/pull/372)
+  [\#372](https://github.com/ocaml-gospel/gospel/pull/372) and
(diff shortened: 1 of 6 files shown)


shym (approved):
I have a question and a suggestion.

- That seems a very natural part of the changes started with #372. Was there a reason we didn’t do it then? In particular, do we expect to break many things with that change? (The commit message could provide a bit of such context)
- I would suggest merging `print_invariant_vs` into `print_invariant`; and, as part of reviewing old but moved code, I would have expected to see `newline` also in `sep` or is that implicitly done somewhere else?

Apart from those small remarks, this looks very good!

n-osborne:
Thanks for the review :-)
I'm not sure why it was not done in the #372 (I think an oversight on my part :sweat_smile: )
It is expect to break Ortac/Wrapper as it reads the invariants. But nothing to hard to fix.

PR #376, opened 2024-01-19
Make type checker save type information

n-osborne:
This PR proposes to marshall the product of the type checker and save it in a file with the `.gospel` file extension so that external tool does not have to run the type checker by themselves.


shym (commented):
This is a simple and nice change, well done!
I’ve pushed in my `drop-ok` branch two extra commits to complete the change of dropping `OK`s and to make `dune` aware of the `.gospel` files that will be generated for succeeding tests. In turn, this makes an extra test that type checking succeeded on documentation :-)

I have one suggestion: it would be nice to teach `gospel check` how to process `.gospel` files, among other things as a way to test the `read_gospel_file` and that the marshalling did work as expected. It would also serve as a simple way to inspect the content of a `.gospel` file.

n-osborne:
Thank you @shym for your comment :-)
I've integrated you two  commits and following your suggestion make `gospel check` handle `.gospel` files and test that reading back the typing information from a `.gospel` file give the same output than typing from the initial `.mli` file.

shym (approved):
This is nice, thank you!
A tiny suggestion: with GNU’s `wc`, `wc -l < bar` is just the line count, so you don’t need an extra cut).
I’m not sure about portability (but that’s already the case with the `cut`, I might argue :-).

mariojppereira (approved):
I believe this pull request is ready to be merged.

n-osborne:
Thanks.
Merging.

Issue #377, opened 2024-01-19
Make the Gospel type-checker modular


shym:
Currently, when the type checker encounters a dependency, it will load and type it on the fly. So common dependencies get type-checked over and over again, uselessly.
It would be nice to modularize the type-checker so that typing a module requires the digested type information of its dependencies instead of retyping them.
This could be done either with dedicated type files (see #376) or trying to piggy-back on current OCaml infrastructure by storing this information in standard files (it would be possible at least in `.cmti` files, by encoding the information as a special global attribute; could be done in `cmi` files?).

The expected benefits would be:
- improved performance,
- easier integration with the standard toolchain, if we manage to have `gospel check abc.mli` require the same files as `ocamlc -c abc.mli`.

shym:
> * gospel type checking can't take any arguments (as it is in a ppx)

It can still grab info from environment variables, I expect, which could give us a way to feed it the information we need.

n-osborne:
Modularizing the type-checker is expected to break how unicity of identifier is guaranteed in the current implementation.
I've raised #381 for discussing this particular topic.

PR #380, opened 2024-01-24
Change `ls_equal` implementation

n-osborne:
This PR proposes to move from physical equality to equality of the identifiers.

This will be needed as we plan to make gospel type checker modular (see #377): different files will be typed by different runs of the gospel program and physical equality won't hold between these runs.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,8 @@
 
 ## Internals
 
+- Use unique identifiers rather than physical equality for `Symbols.ls_equal`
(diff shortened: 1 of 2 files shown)


shym (approved):
LGTM for the code!
I would suggest rephrasing the changelog entry to qualify the new impl(e)mentation.
What about: “Use unique identifiers rather than physical equality for `Symbols.ls_equal`”?

n-osborne:
Agree.
Thanks :-)

PR #386, opened 2024-03-07
Fix `is_ts_tuple`

n-osborne:
This PR proposes to change the implementation of the `Ttypes.is_ts_tuple` function so that it doesnt confuse `unit` for a tuple of length 0.

This has the immediate benefit of concentrating the logic in the function rather than having to check that a type symbol is not `ts_unit` before calling `is_ts_tuple` (hence the simplifcation of the condition in `Typing.process_val_spec`.

This also improves the state of [ortac#207](https://github.com/ocaml-gospel/ortac/issues/207)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,8 @@
 
 ## Internals
 
+- Fix the `is_ts_tuple` function so that it doesn't take `unit` to be a tuple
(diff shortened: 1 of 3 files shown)


shym (approved):
Simple and to the point, LGTM!

PR #391, opened 2024-03-30
Bool

mrjazzybread:
Hello

This pull request changes the Gospel typed AST so that the types of terms are represented with `ty` instead of `ty option`, thereby eliminating the distinction between terms that evaluate to `bool` and propositions. This also means that programs are no longer transformed to turn bools into props. A few caveats: 

- There are still some functions that I want to refactor/rethink their usefulness.
- Predicates are now just syntatic sugar for functions that return bools. Should I remove them from the parser now or should I do another pr?
- Some of the tests didn't pass initially: a test about the `variant` clause didn't pass because now the error message is different. The ill typed program was `variant i = 0`. The old error message was "A term was expected" and the new error message is now "this term has type bool but was expected a value of type integer". In my opinion this message is a bit more clear, but I'm wondering what you think. Another test that failed was one where a predicate is used as an argument for a higher order function that receives a function of type `'a -> bool`. This is now valid under the new typechecker, I suppose this is fine? Another test that failed was the location test, but that is only because it dumps the entire ast, so any changes are always gonna effect it. (side note: this test seems very fragile, may I change it in a future commit so it uses grep and only prints the locations?)
Edit: I just realised that I also have the commits for the path pr here :|, not sure if I should do another pr or if we should wrap up the identifiers one first.

### DIFF
--- a/src/tterm_printer.ml
+++ b/src/tterm_printer.ml
@@ -18,7 +18,7 @@ let print_vs fmt { vs_name; vs_ty } =
   pp fmt "@[%a:%a@]" Ident.pp_simpl vs_name print_ty vs_ty
 
 let print_ls_decl fmt { ls_name; ls_args; ls_value; _ } =
-  let is_func = Option.is_some ls_value in
+  let is_func = ty_equal ls_value ty_bool in
   let print_unnamed_arg fmt ty = pp fmt "(_:%a)" print_ty ty in
   pp fmt "%s %a %a%s%a"
     (if is_func then "function" else "predicate")
@@ -26,7 +26,7 @@ let print_ls_decl fmt { ls_name; ls_args; ls_value; _ } =
     (list ~sep:sp print_unnamed_arg)
     ls_args
     (if is_func then " : " else "")
-    (option print_ty) ls_value
+    print_ty ls_value
 
 let print_ls_nm fmt { ls_name; _ } = pp fmt "%a" Ident.pp_simpl ls_name
 let protect_on x s = if x then "(" ^^ s ^^ ")" else s
--- a/src/tterm_printer.ml
+++ b/src/tterm_printer.ml
@@ -66,9 +66,7 @@ let print_quantifier fmt = function
 
 (* TODO use pretty printer from why3 *)
 let rec print_term fmt { t_node; t_ty; t_attrs; _ } =
-  let print_ty fmt ty =
-    match ty with None -> pp fmt ":prop" | Some ty -> pp fmt ":%a" print_ty ty
-  in
+  let print_ty fmt ty = pp fmt ":%a" print_ty ty in
   let print_t_node fmt t_node =
     match t_node with
     | Tconst c -> pp fmt "%a%a" Opprintast.constant c print_ty t_ty
@@ -76,7 +74,7 @@ let rec print_term fmt { t_node; t_ty; t_attrs; _ } =
     | Tfalse -> pp fmt "false%a" print_ty t_ty
     | Tvar vs ->
         pp fmt "%a" print_vs vs;
-        assert (vs.vs_ty = Option.get t_ty) (* TODO remove this *)
+        assert (vs.vs_ty = t_ty) (* TODO remove this *)
     | Tapp (ls, [ x1; x2 ]) when Identifier.is_infix ls.ls_name.id_str ->
         let op_nm =
           match String.split_on_char ' ' ls.ls_name.id_str with
--- a/src/tterm_printer.ml
+++ b/src/tterm_printer.ml
@@ -18,15 +18,13 @@ let print_vs fmt { vs_name; vs_ty } =
   pp fmt "@[%a:%a@]" Ident.pp_simpl vs_name print_ty vs_ty
 
 let print_ls_decl fmt { ls_name; ls_args; ls_value; _ } =
-  let is_func = ty_equal ls_value ty_bool in
+  let is_pred = ty_equal ls_value ty_bool in
   let print_unnamed_arg fmt ty = pp fmt "(_:%a)" print_ty ty in
-  pp fmt "%s %a %a%s%a"
-    (if is_func then "function" else "predicate")
+  pp fmt "%s %a %a:%a"
+    (if is_pred then "predicate" else "function")
     Ident.pp_simpl ls_name
     (list ~sep:sp print_unnamed_arg)
-    ls_args
-    (if is_func then " : " else "")
-    print_ty ls_value
+    ls_args print_ty ls_value
 
 let print_ls_nm fmt { ls_name; _ } = pp fmt "%a" Ident.pp_simpl ls_name
 let protect_on x s = if x then "(" ^^ s ^^ ")" else s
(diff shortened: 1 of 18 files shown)


n-osborne:
Thanks!

> • There are still some functions that I want to refactor/rethink their usefulness.     

Feel free do do so, some code base cleanup is always welcome and one of the points of this modification is to make the type-checker simpler.

> • Predicates are now just syntatic sugar for functions that return bools. Should I     
  remove them from the parser now or should I do another pr?                             

As stated in #393, there are other problems in the preprocessor, so maybe we can have another PR dealing with syntax. Also, I'm not sure we are completely decided about removing `predicate` as a keyword, but we do want to remove the `bool/prop` disctinction.

>   • Some of the tests didn't pass initially: a test about the  variant  clause didn't    
> pass because now the error message is different. The ill typed program was  variant i =
> 0 . The old error message was "A term was expected" and the new error message is now   
> "this term has type bool but was expected a value of type integer". In my opinion this 
> message is a bit more clear, but I'm wondering what you think. 

This is indeed a nice improvement in the error message!

> Another test that failed
> was one where a predicate is used as an argument for a higher order function that      
> receives a function of type  'a -> bool . This is now valid under the new typechecker, I
> suppose this is fine?

That makes sense and I believe this makes gospel a bit more natural from a programmer point of view :-)

> Another test that failed was the location test, but that is only 
> because it dumps the entire ast, so any changes are always gonna effect it. (side note:
> this test seems very fragile, may I change it in a future commit so it uses grep and   
> only prints the locations?) 

This was to be expected but your idea of using `grep` is very nice! (As `grep` seems to be available on all the CI platform.) Though we may want to keep this PR focused on removing the `bool/prop` disctinction. This can be done in another PR that focuses on cleaning up tests. You can add a short explanation why these tests changed in the commit adding them if you want. (this is really not a strong opinion, feel free to add the `grep` filter if the test modification noise annoys you too much :-) )

> Edit: I just realised that I also have the commits for the path pr here :|, not sure if
> I should do another pr or if we should wrap up the identifiers one first.              

No need to make another PR! This one is already great!
There are two possible situations here:
1. if the modifications in PR 2 depends on the ones done in PR 1, it is natural to base PR 2 on PR 1 (to avoid too many conficts or to use the new modifications)
2. if there is no dependencies, there is no need for PR 2 to be based on PR 1 (although, it is not a big problem, we will just have to be careful in the order in which we merge them).

If in situation `2.` and you want to correct the situation, a `git rebase --onto=UPSTREAM/main @^^` should do the trick (UPSTREAM being the name you gave to the remote `ocaml-gospel` repo).

mrjazzybread:
Hello

There are some functions in the `Tast_helper` module that perform checks that are already performed by the typechecker. For example, in `mk_function` we perform a series of type checking validations that are already done by the typechecker (for example, checking if the type of the returned value is equal to the function's return type). I am either removing these extra verfications or placing them in the typechecker itself whenever is necessary, this way we keep all the typechecking logic in the same place.

mrjazzybread:
Ok, I think this is all I want to do. Besides the unification of the type of booleans and propositions, I need some refactoring, which consisted mostly of removing redundant checks. One of them was checking if a function's spec had no undefined variables. I removed the check, but kept the function that collects all the free variables since it may be useful at some point down the line.

n-osborne:
There are some conflicts that need to be resolved. This would be a shame to lose the benefit of this PR!

mrjazzybread:
Hello

I have resolved the conflicts, but one of the checks failed, specifically with regards to the changelog. I'm not sure what I did wrong :(

n-osborne:
Thanks!
This is the changelog check. There is no new entry in `CHANGES.md` compared to main.
I'm more confused by the presence of commits from #396.

mrjazzybread:
Yes I had already mentioned that I accidentally had the commits for the identifiers PR in this one.

> Edit: I just realised that I also have the commits for the path pr here :|, not sure if I should do another pr or if we should wrap up the identifiers one first.

And you said

> No need to make another PR! This one is already great!

But yes, it is a bit awkward :| As for the changelog, I will add an entry to it ASAP.

n-osborne:
But I believe a rebase on main is necessary as the PR seems to remove some changes done by previous ones.

n-osborne:
For review purpose I've rebased the PR on main [there](https://github.com/n-osborne/gospel/tree/bool-rebase).
You can take it if you want (and if I didn't miss anything) :-)
I believe that if you fetch my branch and then reset yours on it I should disappear from the authors of the commits according to github (which I'm totally ok with).

Apart from the rebasing, I've allowed my self to run ocamlformat on each commit and reformat the commit messages.
Some of the commit messages could use some some love (for example, `Refactoring` is a bit fuzzy).

I've noticed that `Dterm.dterm`s still has an optional type (where `None` means `bool` IIUC). Is there a reason why or could we also remove this?

@mariojppereira do you still want to review this PR?

mrjazzybread:
Ok, I reset the branch with the commits from your rebase, I think everything should be in order. I'll try and rename some of the commit messages tomorrow. Regarding `Dterm.dterm`, the reason why it is an `option` is because it is a late initialized variable that is defined during type unification. When we generate the final typed `term` from a `dterm` we may assume that all `dtypes` are different from `None`. In fact, during the typing phase, when I access the type of a `dterm` I just use `Option.get`. I tried to refactor the code to remove it, however, it requires a lot of modifications that would distract from the main point of the PR.

src/tterm_printer.ml
```diff
@@ -18,15 +18,15 @@ let print_vs fmt { vs_name; vs_ty } =
   pp fmt "@[%a:%a@]" Ident.pp_simpl vs_name print_ty vs_ty
 
 let print_ls_decl fmt { ls_name; ls_args; ls_value; _ } =
-  let is_func = Option.is_some ls_value in
+  let is_func = ty_equal ls_value ty_bool in
```
n-osborne, line 21: Sorry for the late comment: isn't that the other way around?

mrjazzybread: Yes, my bad :( I've updated the code with the fixed version. Just a note: this gospel code:
`(*@ type t = {x : int -> int} *)`

Produces the following result in the pretty printer

``` 
(*@ type t = {x:int -> int}
      function constr#t (_:int -> int)t
      function x (_:t)int -> int *)```

Is this what we want? I don't think we should change this in this PR, but this behaviour seems very unusual.

n-osborne: I believe we should always print the `:` as predicates know also have a type.

The pretty-printer is not really pretty, it is not even a legal printer. I believe it was first written for debugging purpose.
I'm in the process of adding a proper pretty-printer: see #407 and https://github.com/n-osborne/gospel/tree/term-pretty-printer.

mrjazzybread: ok, the git history and the `:` are fixed

n-osborne: Thanks!

n-osborne:
CI failing: you should run ocamlformat on the penultimate commit.

n-osborne:
I've just spotted that the changelog is still update in the 0.3.0 section.
0.3.0 has been released, this PR should go in `Unreleased` section

n-osborne:
Approved and CI happy, merging.
Thanks again!

Issue #393, opened 2024-04-02
Pre-processor bug


ionchirica:
Hi,

When the pre-processor tries to match Gospel keywords, it does not take into account 
white spaces afterwards and so specifications such as:
```ocaml
val function_example : int -> int
(*@ function_example 0 *)
```
Are not accepted by Gospel as it thinks that `function_example` matches the `function` keyword.

n-osborne:
Thanks! Good catch!

Feel free to make a PR to add this test case to the `issue` directory (or even to fix it if you feel like it).
Note that when fixed, there will be a second syntax error in your example as you give a constant in the gospel header.

A bit off topic but as it is on the same line of code: this made me look again at the `pps.mll` file and I've spotted that there is the `val` keyword for introducing ghost declaration. I believe this is a relic of a previous version of the language as it is used in the `vocal` tests and in just one example from the `walkthroughs/union-find.md` documentation file but not really documented as a keyword introducing a ghost value.

I'm not sure we want to keep it, as there is no appearent distinction with `function` and we're already wondering whether to keep `predicate` in #391.

n-osborne:
Closed by #394

PR #394, opened 2024-04-03
Preprocessor bug

ionchirica:
This simple PR aims to fix the issue #393.

I added a test that covers the `function` keyword, I think it suffices but let me know if more are needed :)

### DIFF
--- a/src/pps.mll
+++ b/src/pps.mll
@@ -322,7 +322,7 @@ and gospel ppf start_pos = parse
       Lexing.new_line lexbuf;
       gospel ppf start_pos lexbuf
     }
-  | ("function" | "type" | "predicate" | "axiom" | "val" | "open" ) as k {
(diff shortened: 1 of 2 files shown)


n-osborne (commented):
Thanks!

The fix is simple and to the point. Perfect.

I wonder though why you've added the test in `test/syntax/allsorts_labels.mli`
Maybe the test would be better placed in its own file? `test/syntax/keyword_as_prefix.mli` for example.

ionchirica:
You are right!

I was trying to piggyback on a test file that was more closely related, but I see that it might not be so clear what it tried to achieve.

n-osborne (approved):

n-osborne:
Do you want to clean up the git history or do I "squash and merge"?

ionchirica:
Squash and merge would do just fine :smiley:

PR #396, opened 2024-04-05
Adding paths to Gospel identifiers

mrjazzybread:
Hello again :)

As I said previously, I was trying to change the commit history so that the previous commits were well formatted, but I guess something broke. All the previous commits are now well formatted. I also merged the commits that added tests and identifiers into one. I also added a field to module records with the name of the module (not just the path of the file) for convinience's sake.

### DIFF
--- a/src/typing.mli
+++ b/src/typing.mli
@@ -9,7 +9,6 @@
 (*                                                                  *)
 (********************************************************************)
 
-open Tast
 open Tmodule
 
 type parse_env
@@ -19,11 +18,6 @@ val penv : string list -> Utils.Sstr.t -> parse_env
 (** `penv load_paths module_nm` creates a `parse_env` for typing a module with
     name `module_nm`. The paths in `load_paths` are to be used when searching
     for modules dependencies. *)
-
-val process_sig_item :
-  parse_env -> module_uc -> Uast.s_signature_item -> module_uc * signature_item
-(** `process_sig_item penv muc s` returns a new module under construction after
-    type checking `s` and the typed signature obtained from `s` and `muc`. *)
 
 val type_sig_item : parse_env -> module_uc -> Uast.s_signature_item -> module_uc
 (** the same as above but it drops the typed signature *)
--- a/src/typing.mli
+++ b/src/typing.mli
@@ -20,4 +20,5 @@ val penv : string list -> Utils.Sstr.t -> parse_env
     for modules dependencies. *)
 
 val type_sig_item : parse_env -> module_uc -> Uast.s_signature_item -> module_uc
-(** the same as above but it drops the typed signature *)
+(** `type_sig_item penv muc s` returns a new module under construction after
+    type checking `s`. *)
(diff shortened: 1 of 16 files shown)


n-osborne:
Thanks!

Don't worry about closing the previous PR :-)

This is begining to look great.

In the following example, the call to `f` in `g`'s specification doesn't have a path.

```sh
$ cat > bob.mli << EOF
> module M : sig
>   (*@ function f (i : integer) : bool *)
> end
> 
> open M
> 
> val g : int -> bool
> (*@ b = g i
>     ensures b = f i *)
> EOF
$ gospel dumpast bob.mli | grep "_name.*"
       { Tast.md_name = Bob.M;
                      { Symbols.ls_name = Bob.M.f;
                      [{ Symbols.vs_name = i;
       { Tast.vd_name = Bob.g; vd_type = int -> bool; vd_prim = [];
             { Symbols.vs_name = i_1;
             { Symbols.vs_name = b;
                     { Symbols.vs_name = i_1;
                     { Symbols.vs_name = b;
                       { Symbols.ls_name = infix =;
                            (Ttypes.Tyvar { Ttypes.tv_name = a_1 }) };
                             (Ttypes.Tyvar { Ttypes.tv_name = a_1 }) }
                             { Symbols.vs_name = b;
                              { Symbols.ls_name = f;
                                    { Symbols.ls_name =
                                          { Symbols.vs_name = i_1;
```

I would have expected it to have one. This would allow to know, from the identifier in a term, where it has been defined. I believe this could lead to a better managment of identifiers in the different translations (why3, cfml, ortac...), and this will be needed for a better integration of gospel in odoc (we don't want to ask odoc to resolve our paths). Do we have reasons not to add it?

On a more general note, we tend to use sentences beginning with an infinitive verb for commit messages and changelog entries. Also I'm not sure we completely understand the message for the second commit without reading the code.

And finally, this PR completely breaks Ortac/QCheckSTM tests (I've not investigated why yet). Hence, I'll postpone merging to after the release of gospel 0.3 (and try to hurry up on that) so that I can release Ortac too. We can break everything afterward.

n-osborne:
Apart from the minor comments I made (some can be ignored if you feel like it), this seems good to be merged.

And also that if it is not already done:

> Edit: just noticed I added a file I was using for some quick tests to the repo, I will remove it later 😞

n-osborne:
I would suggest a `git rebase -i aeb411` and removing the `notes` commit :-)
You can also reword the last commit to "Update Changelog" if you will in the same rebase.

mrjazzybread:
Ok I think that's it! I had to make a few more adjustments changes since I added some notes files that should only exist in my fork. It should be all good now.

src/typing.mli
```diff
@@ -20,10 +19,6 @@ val penv : string list -> Utils.Sstr.t -> parse_env
...
     for modules dependencies. *)
 
-val process_sig_item :
-  parse_env -> module_uc -> Uast.s_signature_item -> module_uc * signature_item
-(** `process_sig_item penv muc s` returns a new module under construction after
-    type checking `s` and the typed signature obtained from `s` and `muc`. *)
-
 val type_sig_item : parse_env -> module_uc -> Uast.s_signature_item -> module_uc
-(** the same as above but it drops the typed signature *)
+(** `type_sig_item penv muc s` returns a new module under construction after
```
n-osborne, line 23: Backtick are not correct OCamldoc syntax, use square brackets instead.

n-osborne (approved):
Apart from the tiny last comment this is good to merge.
Thank you!

PR #397, opened 2024-04-09
Use `dune-build-info` for version display

n-osborne:
I believe this is the last PR before Gospel.0.3.0


n-osborne:
CI is green and this is trivial changes that only affects how the `--version` option works.
Merging.

PR #407, opened 2024-05-24
Fix printing of symbols of fixity

n-osborne:
This PR proposes a fix of the way symbols with fixity are printed in `Tterm_printer.print_term`.

This is a first step towards a proprer pretty printer.

This also modifies how the fixity is encoded, proposing to store the information in a field of the identifiers with a sum type rather than in the name itself.


n-osborne:
Ready for review.

n-osborne:
Need rebasing since `Symbols.lsymbol` has been changed by #420 (4a864b1)

n-osborne:
Stale and implemented elsewhere after the type checker rewriting.

PR #418, opened 2024-07-23
Fix incomplete record error

### LINKED ISSUE #417: Incomplete record pattern don't use OCaml syntax
Given a type:

```ocaml
(*@ type t = {a: integer; b: integer}
```

In the current state of Gospel, one can write:

```ocaml
(*@ predicate p (x : t) =
    match x with
    | { a } -> a = 42
```

On the equivalent OCaml code, OCaml compiler would report some missing fields.

Also, Gospel fail with a syntax error if we use an OCaml-correct syntax:

```ocaml
(*@ predicate p (x : t) =
    match x with
    { a; _ } -> a = 42 *)
```

I propose to bring Gospel closer to OCaml.
I already have some experiments in this direction.

n-osborne:
This PR proposes to bring Gospel a bit closer to OCaml regarding incomplete records as value and as pattern. From OCaml compiler, it borrows internally the constructor name for errors, and the error message externally.

It is based on #391, please consider only the last 6 commits.

This PR doesn't bring support for the `{x; _ }` pattern syntax. This is postponed to a future PR as it needs changes in the internal representation of the record pattern, and probably the parser.

Fixes #417


n-osborne:
CI happy, merging

PR #420, opened 2024-08-02
Fix typing of expression with inlined records

n-osborne:
This PR proposes to improve the state of the type-checker regarding expressions involving inlined records.

This work is based on previous, not yet merged, PRs, please consider only the last 12 commits.

The PR is organised in four steps:

- the first four commits add test cases, demonstrating the behaviour we propose to improve
- the next two commits add errors, borrowed from the OCaml compiler so that error messages are closer to OCaml ones
- the next four commits refactor the encoding of symbols, proposing:
  1. to rely on types rather than on boolean encoding to distinguish between fields, functions and constructors
  2. to encode whether a constructor expect a record argument or a tuple one (inspired by how OCaml deals with inlined records)
- the penultimate commit uses the new encoding of symbols to raise the proper errors when needed and build the correct value when possible.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Improved
 
+- Fix typing of expression with inlined record
+  [#420] (https://github.com/ocaml-gospel/gospel/pull/420)
 - Improve error message for unbound record fields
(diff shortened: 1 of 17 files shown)


n-osborne:
CI is happy. Merging.

PR #423, opened 2024-10-20
Stdlib

mrjazzybread:
Here is my pull request for the updated standard library. The goals of
this update are to:

-   Give a precise specification to the defintions within the library so
    that there is no ambiguity as to their meaning

-   Gear this library to the definition of mathematical and logical
    formulae that will be used within specifications. This means removing
    elements such as references and exceptions which do not make sense
    in purely logical contexts

In order to achieve these two goals, we have removed or modified some
of the definitions within this library. I will now go over them and
justify them (note that some of these definitions are still in the
standard library but only so that the tests still pass. In the
future, these should be modified so as to not rely on these
definitions):

<a id="org3638748"></a>

# Removed Definitions

<a id="orgd042fed"></a>

## <a id="orgfb80eab"></a>References

Self explanatory: it does not make sense to talk about references
within purely logical formulae.

<a id="org4f463df"></a>

## <a id="orge05b562"></a>Bitwise Operations

Although one could imagine a specification where one uses bitwise
operations, these operations are out of place in a mathematical
library. Additionally, these definitions are already present in the
OCaml standard library and it does not make sense to duplicate them
here. A better alternative would be to give those functions a gospel
specification and allow users to use them if necessary in their
specifications.

<a id="org6cfd89f"></a>

## Operations on Pairs

Similar justification to [bitwise operations](#orge05b562).

<a id="org4fbd983"></a>

## Sequence Module

In the sequence module we remove the functions `filter_map`,
`fold_right`, and renamed `fold_left` to `fold` In the case of `filter_map`
and `fold_right`, these are just compositions of previously defined
functions; in the case of the former `filter` and `map`, and in the case
of the later `fold` and `rev`. The renaming of `fold_left` felt natural
since there is only one `fold` function. I don't feel very strongly
about these changes and if enough of a need arises, we could add them
again.

<a id="orgfdb17d7"></a>

## List Module

The entire `List` module was removed from the standard library. The main
reason is that the `list` type serves a similar purpose to `sequence`
within specifications and it feels unnatural to have two very similar
ways of reasoning over an ordered sequence of elements. Currently, to
reason over lists in Gospel specifications, we just use a coercion
that transforms them into `sequences`; in the future when we add some
form of representation predicates to Gospel, we could have it so that
the logical model of the type `list` is `sequence`.

<a id="orgd5a11b3"></a>

## Array Module

As i said with regards to [references](#orgfb80eab), it does not make sense to talk
about arrays in a logical context

<a id="org10f0c05"></a>

## <a id="orgc200ac4"></a>Bag Module

The `occurrences` function has been renamed to `multiplicity`. This also
implied changing some of the tests.

The `is_empty` function has been removed since one could simple state
`b = empty` as opposed to `is_empty b`.

The `choose` and `choose_opt` functions have also been removed. It is
unclear why one would use these in a specification.

The `map` has also been removed. This is because it is very difficult to
define it in the case that the parameterized function is not
injective.

The `fold` function has also been removed. This is because it is
impossible to define this function without also defining a canonical
ordering of the elements within the bag.

The `_exists` and `for_all` predicates have also been removed as well as
the `partition` function. To be honest, I do not remember the reasons
why.

The conversion function to a `sequence` has also been removed since this
would also imply a canonical ordering. The conversion from a `sequence`
has also been removed, but I think that might have been a mistake.

<a id="org10270a3"></a>

## Set Module

The `compare` function has been removed, no idea what kind of
specification one could give it or why one would use it in a
specification.

The remaining removed functions are similar to those from the [Bag](#orgc200ac4)
module and therefore I don't think there is a need to explain them
again.

<a id="org5e9e5e1"></a>

## Order module

Since this module is about mathematical definitions it feels out of
place to define a class of valid ordering functions for OCaml
code. Although it makes sense to have this definition somewhere in
Gospel, we should place it somewhere else.

<a id="org568e045"></a>

## System Variables and Exceptions

Similar Reasoning as to [references](#orgfb80eab).

<a id="org0dfdce2"></a>

# Added Definitions

I don't think anything that we have added merits in depth explanation:
mostly it was just defining all the functions we could and
axiomatizing the rest. The majority of these definitions have been
formalized and verified in Coq with a few exceptions these beings some
axioms regarding cardinality and a few recently added axioms regarding
subsequences.

<a id="org2831279"></a>

# Future Work

One glaring omission is the absence of axioms regarding arithmetic operations. These are absent since they are very cumbersome although fairly straight forward.

I would like have better notation for all of these Gospel data types:
notably it would be pleasant to have pretty notation for the `init` and `mem` functions.

Naturally, if these changes are accepted, the tests should be changed
so as to not rely on deprecated definitions. Personally, I think we
should leave them as is and tackle them once we have ways of talking
about ownership of OCaml values. This is because some of these
conflicts, notably with references and arrays, are due to the fact
that Gospel does not have a good way of talking about them in
specifications. I believe that once we have added some form of
representation predicates, that these tests can be rewritten quite
naturally.

Finally, at some point we should separate these modules into distinct
files. For now, since the standard library is still quite small, it is
still manageable, however, as we add more things to it, it can very
quickly become a big mess.


n-osborne (commented):
Thanks!

I suggest you run `git rebase -i --exec "dune runtest --auto-promote; dune fmt" main`. This will put the PR in a cleaner state, render the last commit unnecessary and move the `ocaml-ci` to green.

On the topic of CI, this PR definitely deserve a changelog!

I see you didn't include the Coq/Rocq proof. I would have been at least curious, and I think it could make sense to include it in the repo.

Regarding commit messages, I suggest we follow [these good practices](https://gist.github.com/luismts/495d982e8c5b1a0ced4a57cf3d93cf60#formatting-rules).
They are pretty common and already followed by a non-negligible subset of the commits of this repo.

Now, regarding Gospel formatting, I believe

In commit `44ec14c`, you introduce a white space in the signature of `integer_of_int` at the level of the argument, this is not consistent with the previous style, nor it is a general modification of style. This also has no logical connection with the topic of the commit.

As a general remark about formatting, I feel we lack consistency. This is not just an aesthetic consideration, a consistent formatting will help the reader.

For example, we have three different formatting styles here:

```ocaml
  (*@ axiom append_elems_right :
     forall s s' i.
      length s <= i < length s + length s' ->
      (s ++ s')[i] = s'[i - length s] *)

  (*@ axiom subseq_l :
       forall s i. 0 <= i < length s -> s[i ..] = s[i .. length s] *)

  (*@ axiom subseq : forall s i i1 i2.
    0 <= i1 <= i < i2 <= length s -> s[i] = (s[i1 .. i2])[i-i1] *)
```

My personal preference goes to the first with:
- quantification on the first line
- one antecedent per line (I personally like the arrow at the beginning of the line)
- conclusion on the last line

Some functions are just declared and their semantic is given in the form of axioms, and other are implemented.
I guess it is to minimize the amount of Coq/Rocq proof?
Still, it is a bit surprising for the reader, letting them wonder about whether this is an interface, an implementation file, something else?
The reader also has to switch between reading logical axiom and implementation while going through the standard library, this is not the most pleasant experience.
Some module documentation with some explanation about the purpose of this standard library and how the semantic of the symbols is defined could do a lot of good.

In order to homogenize how the semantic of functions are defined, maybe we could just declare the one that are, at the moment of this review, implemented, and use the implementation term in an axiom? Would this generate too much rework on the Coq/Rocq side?

For example:

```ocaml
  (*@ function singleton (x: 'a) : 'a t = init 1 (fun _ -> x) *)
```

Would become:

```ocaml
  (*@ function singleton (x: 'a) : 'a t *)

  (*@ axiom singletion_definition :
       forall x.
       singletion x = init 1 (fun _ -> x) *)
  ```

This would also have the benefit of letting tools needing an implementation (*i.e* Ortac) free of choosing an efficient one.

For example, the following implementation of `Sequence.mem` is clearly not what we want in the context of dynamic verification:

```ocaml
  (*@ predicate mem (x: 'a) (s: 'a t) = multiplicity x s > 0 *)
```

I'm still not convinced about using the arrow type for `maps`. But I guess this is a design question and it has been already discussed elsewhere.

On the topic of the removal of `fold_right`, I think we should keep it. `fold_left` and `fold_right` are pretty common combinators. I suspect the user discovering a `fold` function would wonder which it is. Also, again regarding dynamic verification (*i.e* Ortac), using a combination of `fold`(`_left`) and `rev` is not what we want for performance reasons. Finally, I think it will render specifications needing `fold_right` a bit cumbersome and heavy. (Note that if you want to get the `fold_right` signature that OCaml programmer's are used to, you'll have to flip the function argument -- and we don't have `Fun.flip`).

The same line of argument applies to the removal of `filter_map`.

mrjazzybread:
I agree with most of your points, I normalized the formatting as you requested. I also made it so each function is abstract and re-added `fold_left` `fold_right`. With regards to the Coq proof, I didn't add it since I only want to put it in the repo once we have a stable version of the library. If there isn't anything too objectionable with what we have here, I will work on adding these changes to the Coq proof and adding it to the PR.

mrjazzybread:
I have added the Coq proof, as well as fixing some small mistakes. There are two Coq files for the proof, the first is the `gospelstdlib_mli.v` which is the automatically generated file which contains a module type `Stdlib` with all of the required definitions and lemmas. The second is the `gospelstdlib_verfied.v` which contains a module `Gospelstdlib` of type `Stdlib`. This means that this project will only compile if all of the definitions in the `Stdlib` module type are present. I have also added a `_CoqProject` file as well as a Coq `Makefile`. Important to note that this proof requires TLC.

In the future, once we integrate my changes into Gospel, we could have an extra test which checks if the `gospelstdlib_verified` file is still congruent with the current standard library by translating it into Coq and checking if the project still compiles.

src/stdlib/gospelstdlib.mli
```diff
@@ -76,56 +76,19 @@
...
-(*@ function (!_) (r: 'a ref) : 'a *)
-(** Reference content access operator. *)
-
 (** {1 Sequences} *)
 
 (*@ function (++) (s s': 'a sequence) : 'a sequence *)
 (** [s ++ s'] is the sequence [s] followed by the sequence [s']. *)
 
 (*@ function ([_]) (s: 'a sequence) (i: integer): 'a *)
+
```
n-osborne, line 91:
```suggestion
```

src/stdlib/gospelstdlib.mli
```diff
@@ -139,222 +102,227 @@ module Sequence : sig
...
-  (*@ type 'a t = 'a list *)
-  (** An alias for ['a list]. *)
+  (*@ axiom init_length :
+        forall n f.
+        n >= 0 ->
+        length (init n f) = n *)
 
-  (*@ function length (l: 'a t) : integer *)
-  (** [length l] is the number of elements of [l]. *)
+  (* TODO : replace init with notation *)
```
n-osborne, line 162: What notation?

src/stdlib/gospelstdlib.mli
```diff
@@ -139,222 +102,227 @@ module Sequence : sig
...
-  (** An alias for the type of arrays. *)
+  (*@ function map (f: 'a -> 'b) (s: 'a t) : 'b t *)
+  (** [map f s] is a sequence whose elements are the elements of [s],
+      transformed by [f]. *)
 
-  (*@ function length (a: 'a t) : integer *)
-  (** [length a] is the number of elements of [a]. *)
+  (*@ axiom map_elems :
+        forall i f s.
+        0 <= i < length s ->
```
n-osborne, line 245:
```suggestion
       in_range s i ->
```

src/stdlib/gospelstdlib.mli
```diff
@@ -139,222 +102,227 @@ module Sequence : sig
...
+  (*@ function set (s: 'a t) (i: integer) (x: 'a): 'a t *)
+  (** [set s i x] returns a new sequence identical to x where the only
+      difference is that the element at index [i] equal [x] *)
 
-  (*@ function sub (a: 'a t) (i len: integer) : 'a t *)
-  (** [sub a pos len] is the array of length [len], containing the elements
-      number [pos] to [pos + len - 1] of array [a]. *)
+  (*@ axiom set_elem :
+        forall s i x.
+        0 <= i < length s ->
```
n-osborne, line 270:
```suggestion
        in_range s i ->
```

src/stdlib/gospelstdlib.mli
```diff
@@ -139,222 +102,227 @@ module Sequence : sig
...
+        (set s i x)[i] = x *)
 
-  (*@ function map (f: 'a -> 'b) (a: 'a t) : 'b t *)
-  (** [map f a] applies function [f] to all the elements of [a], and builds an
-      array with the results returned by [f] *)
+  (*@ axiom set_elem_other :
+        forall s i1 i2 x.
+        i1 <> i2 ->
+        0 <= i1 < length s ->
```
n-osborne, line 276:
```suggestion
        in_range s i1 ->
```

src/stdlib/gospelstdlib.mli
```diff
@@ -139,222 +102,227 @@ module Sequence : sig
...
+        0 <= i2 < length s ->
```
n-osborne, line 277:
```suggestion
        in_range s i2 ->
```

src/stdlib/gospelstdlib.mli
```diff
@@ -139,222 +102,227 @@ module Sequence : sig
...
+  (*@ axiom rev_length :
+        forall s.
+        length s = length (rev s) *)
 
-  (*@ function fold_right (f: 'b -> 'a -> 'a) (a: 'b t) (init: 'a) : 'a *)
-  (** [fold_right f a init] is [f a[0] (f a[1] ( ... (f a[n-1] init) ...))],
-      where [n] is the length of [a]. *)
+  (*@ axiom rev_elems :
+        forall i s.
+        0 <= i < length s ->
```
n-osborne, line 290:
```suggestion
        in_range s i ->
```

src/stdlib/gospelstdlib.mli
```diff
@@ -139,222 +102,227 @@ module Sequence : sig
...
-      builds an array with the results returned by [f]. *)
+  (*@ axiom extensionality :
+        forall s1 s2.
+        length s1 = length s2 ->
+        (forall i. 0 <= i < length s1 -> s1[i] = s2[i]) ->
+        s1 = s2 *)
 
-  (*@ predicate for_all (f: 'a -> bool) (a: 'a t) *)
-  (** [for_all f a] holds iff all elements of [a] satisfy the predicate [f]. *)
+  (*@ function fold_left (f: 'a -> 'b -> 'a) (acc: 'a) (s: 'b sequence) : 'a *)
```
n-osborne, line 299:
```suggestion
  (*@ function fold_left (f: 'a -> 'b -> 'a) (acc: 'a) (s: 'b t) : 'a *)
```

n-osborne (commented):
Thanks!

I went through the `Sequence` module again and made some inlined suggestions (feel free to squash them at the right place afterward).
I also have two more general remarks:

1. Not all functions and predicates have an informal documentation
2. I wonder about the order. Some axioms at the beginning of the module make use of functions that are defined later (`append_*` axioms mostly)

src/stdlib/gospelstdlib.mli
```diff
@@ -139,222 +102,227 @@ module Sequence : sig
...
-end
+  (*@ function empty : 'a t *)
+  (** [empty] is the empty sequence. *)
 
-(** Lists
+  (*@ axiom empty_length : length empty = 0 *)
 
-    The type ['a list] and the constructors [[]] and [(::)] are built-in. *)
+  (*@ function init (n: integer) (f: integer -> 'a) : 'a t *)
+  (** [init n f] is the sequence containing [f 0], [f 1], [...] , [f n]. *)
```
n-osborne, line 155:
```suggestion
  (** [init n f] is the sequence containing [f 0], [f 1], [...] , [f (n - 1) ]. *)
```

n-osborne:
Since we have partial functions in Gospel, I am wondering what is the semantic of second order functions when one of the function argument is partial.

For example, the `Sequence.init` function is defined as:

```
(*@ function init (n: integer) (f: integer -> 'a) : 'a t *)
```

My guess would be that `init n f` is not defined when `f` is not defined for an `x` in `0..n-1]`.

Do we need to document that? And if we do, how do we do it?

mrjazzybread:
Hello, I fixed all of the things you noted. I also updated the change
log. With regards to what you noted about higher order functions, I
don't understand the problem. When you say `init` is undefined when `f` is
not defined in the given range, wouldn't this problem arise regardless
of whether or not the function was partially applied?

In any case, I would say that if Gospel assumes functional
extensionality then I would say that we do not need any special
specifications for partially applied functions since we can leverage
the fact that `forall f g x y. f x = g <-> f x y = g y`.

n-osborne:
Thanks!

The more I review this PR, the more I think this is a very nice new standard library.

I haven't reviewed the Coq/Rocq proof, as I am certainly not an expert.
Am I correct in understanding that this provides a Coq/Rocq implementation of the Gospel standard library proven correct *w.r.t* the Gospel axioms?

I have just a few remaining small remarks:

- line 347: `Creates` -> `creates`
- move line 428 to 422, attaching the informal documentation to its item
- line 466: predicate `finite` doesn't have any informal documentation (even a trivial one)
- line 510: maybe we can remove it already?
- line 666: predicate `finite` doesn't have any informal documentation (again, even a trivial one)
- line 673: maybe we can remove it already?

> With regards to what you noted about higher order functions, I don't understand the problem.

There is no problem :-), I was just wondering. In retrospect, this is something that should be documented in the language definition rather than in the standard library documentation.

How do you want to proceed about the three modules at the end to be removed? You want to do it in this PR or another one? I believe this should be done before a release with the new standard library.

n-osborne:
> The same line of argument applies to the removal of filter_map.

I believe you didn't add back `Sequence.filter_map`. We do use it in a test in Ortac ([here](https://github.com/ocaml-gospel/ortac/blob/main/plugins/qcheck-stm/test/hashtbl.mli#L42) and [there](https://github.com/ocaml-gospel/ortac/blob/main/plugins/qcheck-stm/test/hashtbl.mli#L67).

n-osborne:
It makes sense to remove them in another commit as it will contain an update of the tests.

Thanks for adding back `Sequence.filter_map`! (If you want to be acribic about your git history, you remove the definition in the first commit ([here](https://github.com/ocaml-gospel/gospel/pull/423/commits/e2e183b70411e55d25ff0642f0b8d135f0b659bb#diff-8364d888288b2071c3f6b0929cd505076b9f94ae2a102b354f0f910f201ec07cL179)) before adding it back in the third one ([here](https://github.com/ocaml-gospel/gospel/pull/423/commits/18d63fbda05bf7413a89a3368cb280e2e3dd3aeb#diff-8364d888288b2071c3f6b0929cd505076b9f94ae2a102b354f0f910f201ec07cR255)), with a better documentation though.
But it is totally ok as it is :smiley: )

mrjazzybread:
I removed the deprecated modules and changed the tests accordingly,
which consisted of coercing values of type `list` and `array` into
`sequence` or something similarly simple.

A few other changes I made:

-   Since the `Order` module was used by several tests, instead of copying
    it into each file, I added an `Order.mli` file with the `is_pre_order`
    predicate.
-   Related to the previous point, I fixed a bug where Gospel could not
    open a module contained in a file that starts with a capital
    letter.
-   I removed the syntax for array access since we want users to use
    sequences instead of reasoning over the array directly.
-   I re-added `fold` in the `Set` module since one of the Vocal examples
    used it and I didn't want to change it too much. I did however,
    change its signature: `fold` now takes two functions, one that maps
    elements into some other type `'b` and a function that combines two
    values of type `'b`. In the specification for `fold`, the latter
    function must be a commutative monoid.
-   To simplify the specification of `fold` I also added a definition for
    a monoid and a commutative monoid in the top level of the library.
-   I changed the history so that filter_map isn't removed

n-osborne:
Excellent: Thank you :-)

I believe [41162f0](https://github.com/ocaml-gospel/gospel/pull/423/commits/41162f0e24b5a0319204712be42932f4436c8171) could have deserved its own PR. But is is ok.

Once the conflict on the changelog has been resolved, this PR can be merged if you are ok with that.

mrjazzybread:
I thought of making it its own PR, but since this one is almost done, I figured why not. The arithmetic operators don't have specifications, but I don't think that is a huge deal breaker. I am pretty satisfied with the current state of this library.

PR #425, opened 2024-10-27
Update to the Documentation Site

mrjazzybread:
Hello

I  was analyzing the Gospel documentation website and as I was exploring, I noticed that there was a small discrepancy: we state that the type checker differentiates between terms and formulae when that is not the case anymore. Additionally, there were some outdated node packages which led to security problems. I figured that these two issues were relevant enough that they warranted a small PR to fix them. I also changed the git hub actions to use an updated version of node.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
         with:
-          node-version: '16.14'
+          node-version: '20.18'
       - name: Test Build
         run: |
           cd docs/
--- a/.github/workflows/localdoc.yml
+++ b/.github/workflows/localdoc.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Set up node
         uses: actions/setup-node@v3
         with:
-          node-version: '16.14'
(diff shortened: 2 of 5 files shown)


n-osborne (commented):
Thanks!

- First commit message: Updated -> Update
- The second commit contains an update of node version while it is about updating the content of the documentation

I haven't investigated the incompatibility reported by the CI when trying to build the documentation.

Apart from that, I obviously agree with the change.

mrjazzybread:
Ok I fixed the issues you pointed out. With regards to  incompatibility, I think if you change the node version on the Gospel side it will work.

n-osborne:
It seems that if you update the `.github/workflows/documentation.yml` the same way you've changed the `.github/workflows/localdoc.yml`, it should work. (I couldn't reproduce though).

Regarding [1b15126](https://github.com/ocaml-gospel/gospel/pull/425/commits/1b151260db7967e332271e88b8771943ad154722), it is cleaner to use `git rebase` in these cases, no need for a merge commit.

mrjazzybread:
Ok, I fixed the merge commit and changed the node version on the `github/workflows/documentation.yml` file.

n-osborne:
Excellent! Thank you!

Issue #431, opened 2024-12-17
Typechecking bug in higher order Gospel functions


mrjazzybread:
As I was updating the Gospel standard library, I came across a strange bug in the Gospel typechecker

If I create a function like

`(*@ function f (f : int -> 'a) : 'a *)`

Then write an axiom

`(*@ axiom fail : forall x. f (fun _ -> x) = x *)`

It fails with the following cryptic error

```
gospel: internal error, uncaught exception:
        Gospel.Ttypes.TypeMismatch(_, _)
```

However, the following axiom is typechecked correctly

`(*@ axiom fail : forall x : 'a.  f (fun _ -> x) = x *)`

n-osborne:
Thanks for reporting this.

What would be the expected behaviour in your opinion? Should the axiom be accepted or should the type-checker should fail with a proper error message?

Apparently, the uncaught exception is raised in `Ttypes.ty_match`.

Failing with a proper error message can simply be achieved by using the `Warning` mechanism.
If we want that example to be accepted by the type-checker, we are lacking a bit of type inference I guess.

Note that the following example is produced using the last released version of Gospel. Line numbers are expected to be different with `main`.

```sh
$ cat <<EOF > foo.mli
> (*@ function f (g : int -> 'a) : 'a *)
> (*@ axiom fail : forall x. f (fun _ -> x) = x *)
> EOF
$ OCAMLRUNPARAM=b gospel check foo.mli
gospel: internal error, uncaught exception:
        Gospel.Ttypes.TypeMismatch(_, _)
        Raised at Gospel__Ttypes.ty_match in file "src/ttypes.ml", line 171, characters 15-58
        Called from Gospel__Tterm_helper.t_app in file "src/tterm_helper.ml", line 148, characters 10-34
        Called from Gospel__Dterm.term in file "src/dterm.ml", line 340, characters 10-54
        Called from Gospel__Dterm.term_node in file "src/dterm.ml", line 365, characters 17-37
        Called from Gospel__Dterm.term in file "src/dterm.ml", line 340, characters 10-54
        Called from Gospel__Dterm.term_node in file "src/dterm.ml", line 396, characters 14-30
        Called from Gospel__Dterm.term in file "src/dterm.ml", line 340, characters 10-54
        Called from Gospel__Dterm.fmla in file "src/dterm.ml" (inlined), line 431, characters 18-34
        Called from Gospel__Typing.fmla in file "src/typing.ml", line 544, characters 11-22
[...]

PR #435, opened 2024-12-28
Fix bug in creation of fresh type variables

mrjazzybread:
Happy Holidays!

I figured out the problem with the bug from [#431](https://github.com/ocaml-gospel/gospel/issues/431): when a fresh type
variable is created, it is "logged" into a hash table, using the
`ty_of_dty` function, that maps it into another type variable with a
unique name. However, there was one instance where we were creating
type variables without adding them to the hash table which was causing
problems when we ran the `ty_match` function (which I believe exists
only for integrity checking). This PR should fix this issue.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,6 +46,8 @@
 
 ## Internals
 
+- Fix bug in creation of fresh type variables
+  [\#435](https://github.com/ocaml-gospel/gospel/pull/435)
 - Fix premature parsing of specification keywords in the preprocessor.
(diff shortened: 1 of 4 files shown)


n-osborne:
Nice catch!

Maybe it could be an idea to add a test showing it works (so that we don't break it again later).

As a side not, the changelog check in the CI seems broken as you didn't add an entry in it but the check passes...

n-osborne (approved):
Excellent! Thanks!

This is merely lacking an entry in the changelog to be merged.

mrjazzybread:
Done!

Issue #430, opened 2024-12-05
Gospel injects `fs_apply` symbols in the typed AST.


n-osborne:
This issue has been raised offline by @mrjazzybread

This behaviour is observed when there is a function application where the term for the function is not a name.

For example, in this function definition:

```ocaml
(*@ function f (b : bool) : bool = (fun x -> x) b *)
```

Gospel type-checker will produce something like `Tapp (fs_apply, [Tlambda ...; Tvar "b"])` for the body of the function.

In fact, it can't produce anything else in the current state of the typed AST as the application node (`Tapp`) in `Tterm.term_node` takes a `lsymbol` as its function argument.

We should make `Tapp` looks like:

```ocaml
  | Tapp of term * term list
```

and go from there.

n-osborne:
@mrjazzybread can you confirm that this is not the case anymore since your rewrite? (so we can close the issue)

mrjazzybread:
It is not an issue anymore, we now have a single constructor for function applications `Tapply`.

n-osborne:
Then I close the issue.
Thanks.

PR #436, opened 2025-01-05
AST node for application of arbitrary terms

mrjazzybread:
Hello

As mentioned in [#430](https://github.com/ocaml-gospel/gospel/issues/430), Gospel injects `fs_apply` when applying function
terms that are not simple names. This pull request adds a dedicated
AST node `Tapply` for arbitrary function terms and changes the name of
the current AST node for named function application to `Tidapp`.


n-osborne (dismissed):
Thank you!

We may want to discuss the modifications to the AST.

n-osborne:
Stale

PR #437, opened 2025-01-06
Fix bug on change log Github action

mrjazzybread:
Hello

This PR fixes the bug in the git action for the change log where
running git diff resulted in a "bad revision" error. This bug was
introduced in [bf39b4b](https://github.com/ocaml-gospel/gospel/commit/bf39b4b1824050409f4ad6b7b4e16b81845324d1) when the version of the `checkout` git hub action
was set to 4.


n-osborne (approved):
Thank you!

PR #438, opened 2025-01-13
Start design document

n-osborne:
This design document at the root of the project is a place where we can discuss design and implementation of the Gospel type-checker.


n-osborne:
@mrjazzybread please feel free to add any remarks you see fit.
The more we put in this PR the less we'll have isolated commits in the future.

PR #441, opened 2025-01-16
Add Lenses

mrjazzybread:
Hello

This PR is the first (I guess second if you count the updated standard
library) batch of changes I am proposing to the Gospel type checker.

<a id="org07bcfe1"></a>

# Lenses

One of the major changes we made to the type checker and the surface
language is *lenses*, lightweight notation to describe ownership of
values and their connection to their logical counterparts.

As an example, let us see how we might specify a queue using lenses:

    type 'a queue
    (*@ model view : 'a sequence *)
    
    val push : 'a queue -> 'a -> unit
    (*@ push q x
        consumes q @ 'a queue
        produces q @ 'a queue
        ensures q.view = cons x (old q.view) *)

The `consumes` clause of this specification states that we receive the
argument `q` with the lens `'a queue`, which means we receive `q` with full
ownership. The `produces` clause states that we return the `q` with the
same lens and therefore the same ownership conditions. Since we have
full ownership of `q`, we can access its model in the specification.

Since receiving and returning the same ownership is a very common use
case, one can use the `modifies` which is desugared into a `produces` and
`consumes`.

    val push : 'a queue -> 'a -> unit
    (*@ push q x
        modifies q @ 'a queue
        ensures q.view = cons x (old q.view) *)

There is one more clause that we added to the language: `preserves`,
which is desugared into a `produces`, `consumes`, as well as marking the
argument as read only. For example, the following specification:

    val length : 'a queue -> int
    (*@ l = length q
        preserves q @ 'a queue
        ensures q.view = Sequence.length (q.view) *)

is desugared into

    val length : 'a queue -> int
    (*@ l = length q
        consumes q @ 'a queue
        produces q @ 'a queue
        ensures q.view = old q.view
        ensures q.view = Sequence.length (q.view) *)

<a id="org09738e6"></a>

### Other Lenses

The only other lens currently available is the `loc` lens, which
describes having no ownership over OCaml values. For example, if we
have a type `t` and we want to write a function for physical equality:

    type t
    (*@ model v : integer *)
    
    val ph_eq : t -> t -> bool
    (*@ b = ph_eq x y
        preserves x @ loc
        preserves y @ loc
        ensures x = y *)

In the postcondition, `x` and `y` are of type `t loc`, meaning that their
model `v` is not accessible. In this case, the term `x = y` is comparing
the physical locations of the two values, not their logical values.

<a id="orga830b58"></a>

### Defaults

When the lens is omitted in a specification, we assume full
ownership. This means we can rewrite the specification for `push` as:

    val push : 'a queue -> 'a -> unit
    (*@ push q x 
        modifies q
        ensures q = cons x (old q)*)

and get the same specification as before. This means that the changes
proposed to the type checker are backwards compatible, although the
representation of values within the AST has changed.

Additionally, if no ownership clause is given for a variable, then we
assume a `preserves` clause with full ownership. This means that the
`length` example could be rewritten as:

    val length : 'a queue -> int
    (*@ l = length q
        ensures l = Sequence.length *)

<a id="org487a112"></a>

## Logical quantifiers

Sometimes, we might wish to state a universal property about some
OCaml type which requires us to use a lens to make . For example, if
we wish to state that all queues are non-empty, we could write.

    (*@ axiom non_empty : 
          forall q : 'a queue. q.view <> empty *)

Which is desugared into

    (*@ axiom non_empty :
          forall q @ 'a queue : 'a queue. q.view <> empty *)

<a id="orgb2b0626"></a>

## Future Work

Lenses are still quite limited, they only allow us to either claim
full or no ownership. In the future, we might want to add something in
between where we state that we own a data structure but not its
contents. For instance, we could rewrite the length example as:

    val length : 'a queue -> int
    (*@ l = length q
        preserves q @ loc queue
        ensures l = Sequence.length q *)

Since we use the `loc queue` lens instead of `'a queue`, `length` only
requires partial ownership of the data structure. This entails that
the type of `q` in the ensures clause is `'a loc sequence`. Although this
is not very difficult to add to the type checker, we have not worked
out what the semantics of this would be.

<a id="orgbc6f48f"></a>

# Unnamed model

When giving a `model` to an OCaml type we can now choose to have it be
unnamed, for example, the previous queue example can be rewritten as:

    type 'a queue
    (*@ mutable model : 'a sequence *)
    
    val push : 'a queue -> 'a -> unit
    (*@ push q x
        modifies q @ 'a sequence
        ensures q = Sequence.cons q x *)

The only details worth noting is that if a type has an unnamed model,
it cannot have any named `model` (rejected by the parser). Also, the
previous syntax for models remains the same, meaning if a user has a
single model, they can still , meaning this is also
backwards compatible

This also means that coercions can finally be removed and replaced
with these unnamed models.

<a id="org1c604dd"></a>

# Suite of examples

I also added a suite of examples for Gospel specifications, which
serve as an integrity check for lenses. Additionally, examples of
Gospel specifications are just a good idea to have in the repository
in general.

<a id="org68466df"></a>

# Default Specifications

The Gospel type checker inject a default type specification into types
and value declaration when one is absent. In the case of functions,
all their arguments are `preserved`. In the case of type specifications,
its model is assumed to be a reflection of its OCaml type and the type
is marked as not ephemeral

<a id="org95c8d3d"></a>

# Removal of the mutable keyword

When defining `model` fields, it is no longer possible to mark one model
field as `mutable` and another as not `mutable`. This is because the
Separation Logic of such a scheme becomes very difficult to define in
the case of having nested mutable data structures.

In the case of a data structure where one of its model fields does not
change (e.g. an array with a constant capacity), we now write:

    type 'a array
    (*@ ephemeral
        model elts : 'a sequence
        model capacity : integer *)

And in each function we state that only capacity is modified

    val set : 'a array -> int -> 'a -> unit
    (*@ set arr i x
        modifies arr.elts
        ensures ...*)

It's a bit more painful to write, but it is much easier to define the
semantics of this specification.

<a id="orgd8b3571"></a>

# Next PR

I still have another PR lined up, which will be for the translation of
Gospel into Separation Logic. I am splitting them in two for the sake
of keeping these PRs focused and manageable.


mrjazzybread:
Hello

Thank you very much for your review, I will go over your points in
more detail later (I have a lot on my plate right now). The rebase I
just pushed is nothing special, just renaming all instances of
"spatial" with "lens". I will go over some of the simpler points you
raised.

<a id="org6c0ba20"></a>

# Default specifications

When we define the Separation Logic semantics of a Gospel annotated
OCaml program, we have to decide how we want to translate an empty
specification. Instead of having the Separation Logic semantics module
decide the meaning of an empty specification, this is done during type
checking. I think this is better since, when translating Gospel to
some verification framework, one might either use the Separation Logic
semantics (e.g. My CFML prototype) or the Gospel typed AST
(e.g. Cameleer). Given this, both of these representations should
agree on what the default specification is when one is omitted. I
could have put this in the following PR, but I wanted that one to just
be the Separation Logic semantics and this PR have all (or at least
most of) the changes to the type checker.

<a id="orgab7a399"></a>

# *Ephemeral* keyword

Regarding the `ephemeral` keyword, maybe we could replace it with
`mutable` since it is no longer used.


<a id="org661ae32"></a>

# Remove mutable keyword `f27597e`

Regarding this commit, do you want me to split the documentation into
another commit? Keep in mind that the code within the docs has to be
changed regardless since the documentation must type check for the
project to compile. If we split the documentation into another commit,
we will have a commit where the docs do not reflect the example
specification (which I suppose is fine).


<a id="orgb185706"></a>

# Documentation

I was pretty lazy about documenting my changes to the Gospel type
checker and I do apologize for that 😐. As soon as I can,I will add
some comments guiding what is happening in `val_process_spec`.

PR #449, opened 2025-03-27
Type checker rewrite

mrjazzybread:
# Table of Contents

By popular demand :slightly_smiling_face: 

We have already discussed this at length offline so I won't talk about
motivation for this pull request or the reasoning for some of the
design decisions. Instead, I'm going to go over some of the ways that
this implementation differs (for the better (mostly)) than the
original.

-   Documentation. It exists now.

-   I have added an `ARCHITECTURE.md` file to the root of the project
    explaining the implementation in a very high level way as well as
    detailing how a Gospel spec flows through the program. This is not a
    replacement for inline documentation for each module explaining what
    it does but rather a way to talk about the relationship between the
    most critical modules. I also added a section explaining the testing
    framework since when I started I had no idea how it worked (although
    in hindsight it is quite sensible) and I figured this might help
    anyone going through what I did.

-   Typing error messages are less specific :slightly_frowning_face: . Less specific because
    unfortunately, if we have situation like `1 + ""` we would want to
    print an error saying "Expected type `integer` but found type `string`",
    however in Inferno, when two types are not unified, we don't know
    which was expected, so we would have to print a message saying
    "Mismatch between type `string` and `integer`". We still have the
    location so we can point the user to where the error happened (check
    the tests to see how the error messages look), but it will be a
    little less explicit about what actually happened.

-   Typing error messages are more specific :slightly_smiling_face: We now have a mechanism
    that traverses the two types and checks the first component that is
    not equal. For example, if we have `Sequence.singleton 0 =
      Sequence.singleton ""`, 
    we will print the message "Mismatch between
    `string sequence` and `integer sequence` Type `string` is incompatible
    with `integer`". Note that this is not specific to Inferno and it
    could have been done in the previous implementation.

-   Generated type variable names are a bit more readable. Before, when
    the Gospel type checker would generate a new name, it would be
    always be `'aN` where `N` would be some number. This would be a bit
    annoying because eventually the `N`'s would get into the hundreds and
    the error messages would become pretty noisy. Additionally, it
    would make translations into Rocq extremely annoying since the names
    of the type variables would change when you added new type
    definitions. Now, names are generated like the `OCaml` compiler does
    where it starts with `'a` then `'b` then `'c`, etc&#x2026; This mechanism is
    strong enough to allow for an arbitrary amount of generated type
    variables (although I haven't tested this fully).

-   The standard library is now part of the compilation
    chain. Previously, if you made a change that messed up the standard
    library, Gospel would not complain but when you tried to type check
    any file it would complain that the standard library did not
    exists. Now the standard library is type checked whenever we compile
    Gospel and is loaded from a marshalled file at run time.

-   Although I haven't changed the testing infrastructure (I made a few
    tweaks to the code but nothing that changed the underlying
    behaviour) I have added my own tests and pretty much abandoned the
    previous ones. Although I have a similar amount of tests that
    current Gospel has (about 200 and something), keep in mind that I
    have only implemented a relatively small fraction of what current
    Gospel has. I wanted to write more but I was already taking too long
    to do this PR :grimacing:.
    
-   If a user writes `M1.M2.x`, these module accesses will still appear in
    the typed tree. I'm not sure why the previous version of Gospel
    didn't do this but it also made translations into Rocq more annoying
    then they had to be. Additionally, when a user binds some variable
    to some type such as `x : M.t1` and `M.t1` is an alias for `t2`, the typed
    AST will contain both the annotation the user wrote (`M.t1`) as well as
    the alias (`t2`).

-   When typechecking Gospel, we still generate a marshalled file, but
    now I insert it into a `_gospel` directory. Because of this, I had to
    increase the OCaml version since the `mk_dir` function only became
    available on `4.14`, surprisingly.

And those are the important bits that I have done. As you can see, these
are A LOT of commits so it is impossible to describe in full detail
what I have done, but these are the highlights.


n-osborne:
Thanks for this work!

I won't be reviewing in details this PR as it consists of almost 80 commits and counting. It is even now combining rewriting the type-checker and adding the lenses. As a consequence I won't take the decision to merge it alone neither.

I believe @mariojppereira is following the work more closely than I.

That being said, I see that the PR is not building for some time now. I can take a look if that helps (there are at least some missing dependencies).

As the code doesn't compile I haven't been able to take it for a spin, but what you describe seems quite an improvement (along side the improvement of the code base).

> -   Although I haven't changed the testing infrastructure (I made a few
>    tweaks to the code but nothing that changed the underlying
>    behaviour) I have added my own tests and pretty much abandoned the
>    previous ones. Although I have a similar amount of tests that
>    current Gospel has (about 200 and something), keep in mind that I
>    have only implemented a relatively small fraction of what current
>    Gospel has. I wanted to write more but I was already taking too long
>    to do this PR :grimacing:.

I have to admit I'm a bit surprised here. Why abandoning existing tests? I'm
also a bit disappointed as @shym and I spend quite some time cleaning up the
tests (#312) in the prospect of an eventual rewriting (which is happening here
and now).

mrjazzybread:
Hello

The PR has been a little stale since I have been in Portugal for the last 2 weeks.  I returned last night so I should get back to it.  A few notes:

Tests: I scrapped the original tests only for convenience: I wanted to test the things that I was implementing as I was implementing them and I couldn't really do that with the original test suite so I started my own.  There's a decent amount of them but not enough in my opinion.  I was going to combine the original Gospel tests with mine once I complete the rewrite so if you guys have improved them, that is really good!

Lenses: these are going to be added in a very limited way initially, I am going to first add them as an internal mechanism for turning an OCaml type into its Gospel model (and we assume full ownership of the OCaml value).  We will then add these as an explicit syntatic construct where users can choose .

Code not compiling: I make sure that all of my commits compile and pass each test, however in my workflow I sometimes have a `wip` commit for larger commits (I am currently working on OCaml function specifications :grimacing: )  I have moved this `wip`commit into another branch so if you want to give the code another try, it should work.

Once OCaml function specifications are added, the rewrite of the typechecker should be basically done (at least it should be able to process everything the old typechecker could) .   Once I am done with this, I will let you know and we can all have a meeting about it.

n-osborne:
It would be nice if you could include this use case in the tests: https://github.com/ocaml-gospel/ortac/issues/304 :-)

mrjazzybread:
Hello

This specific case is a little problematic since it is specifying an
OCaml higher order function and we are still not sure how those should
be handled.  In the example shown, the specification is only valid if
`f` has no side effects, otherwise it is not allowed to be used 
directly within Gospel specifications.

If we introduce some annotation to state that a function is pure (or
just assume it by default), then we could map each argument and return
type to their model (in this case, `int -> int` would be `integer ->
integer`).  
Nevertheless, at the moment I assume that OCaml functions
do not have a valid logical representation and therefore are unusable
in specifications.  In my next meeting with my advisors I will bring this
up to know how they feel about directly using OCaml functions in 
specifications.

jmid:
> If we introduce some annotation to state that a function is pure (or just assume it by default), then [...]

As I noticed and commented on in ocaml-gospel/ortac#270 this default is already the case for the current version of Gospel. As such, I believe removing support for pure first-class functions would be regression.

n-osborne:
> This specific case is a little problematic since it is specifying an OCaml higher order function and we are still not sure how those should be handled. In the example shown, the specification is only valid if `f` has no side effects, otherwise it is not allowed to be used directly within Gospel specifications.

This is not problematic in the context of Ortac. I understand that in a proof setting, you'll have a proof obligation about the purity of the functional argument in order to be able to use this specification in a call to the higher-order function. In the context of Ortac/QCheck-STM this obligation is discharged by the fact that QCheck only generates pure functions.

Building on @jmid's comment, I strongly think that we should keep a simple and straightforward way to talk about higher-order functions expecting a pure functional argument (the current way). This is the most common case. Also, for now the experiments on talking about non-pure functional arguments involve quite some setting that we don't want to use to talk about pure functional ones.

mrjazzybread:
Hello

I am very nearly done with this rewrite :partying\_face: Just wanted to
write some more updates:

-   Value specifications are supported.  The syntax is fairly different
    however, we had discussed this a few months ago in the last type
    checker meeting before the rewrite.  Basically, specifications are
    written so that `requires`, `modifies`, `consumes`, `diverges` and `pure`
    annotations are written first, then we have a `let` to bind the return
    variables and the arguments.  This is then followed by a series of
    `ensures` clauses.  In the case of exceptional post conditions, the
    user can write `match ... with` and match over the corresponding
    exceptional cases.

-   When we state `modifies x` (or any ownership clause), we check if `x` is
    a mutable type.  If not, we raise an error. A variable `x` is
    considered mutable if it uses any type names that are mutable, with
    the caveat that all arrow types are considered immutable.

-   OCaml names cannot be used in Gospel terms.  Some exceptions:
    -   OCaml type variables can be used in Gospel terms, although the
        semantics of this are still a bit unsatisfactory.
    
    -   When we have an OCaml type with a named Gospel model, a Gospel
        record type is added with the same fields.  For example:
        
            type 'a array
            (*@ mutable
                model contents : 'a sequence
                model length : integer *)
        
        Is sugar for:
        
            (*@ type 'a array = { contents : 'a sequence; length : integer } *)
            
            type 'a array
            (*@ mutable model : 'a array *)
        
        This means that the `array` type is available in Gospel, but we can
        only access the `contents` and `length` field.  Another consequence is
        that we are not allowed to state that only the `contents` is
        mutable and not the `length`.
    
    -   OCaml variables can be used within the specification, but only as
        their models.  For example:
        
            val length : int array -> int
            (*@ let r = add arr in
                  ensures x.length = r *)
        
        In this specification (assuming the `model` of `int` is `integer`), `arr`
        is of type `integer array` (where `array` is the `Gospel` record type)
        and `r` is of type `integer`.  If we use in a specification a value of
        a type that does not have a Gospel `model`, we raise an exception.
    
    -   OCaml top level values can be used within specifications but only
        if they are listed in an ownership clause.  For example.
        
            val top_array : int array
            
            val top_length : unit -> int
            (*@ preserves top_array
                  let r = top_length n in
                  ensures r = top_array.length *)
        
        Is valid, however, if we were to remove the `preserves` clause, this
        would raise an exception.  As before, the type of `top_array` within
        the Gospel specification corresponds to its `model`.

Some relevant things that we still haven't done:

-   Pattern matching.  We are trying to go without it and for the most
    part it is doable, however we still have the issue of how to work
    with the `option` type.

-   Module types and Functors.  The former should be easy, the later
    might be tricky.  I don't know if there are a lot of Ortac specs
    that use functors, but if it's negligible, we could merge without
    them.

-   Lenses.  I want to get the type checker to the same point it was
    before this rewrite so lenses are going to come afterwards.  I wrote
    the code with this in mind so the way that ownership clauses are
    processed is a little convoluted considering the current simplicity
    of the system, however, in the future it will make it easier for me
    to integrate lenses.  Another consequence of this is that in the AST
    we have `produces` clauses but these are not available in the surface
    syntax.  This because the `produces` clause is only useful if you have
    lenses.

There's a lot more to this pull request, but they are either details
or things that we have already discussed. I wanted to take this
opportunity to ask if you could send me the cleaned up tests you
talked about a few weeks ago.  Also, feel free to complain if you have
any problems with the implementation/organisation.

n-osborne:
> I am very nearly done with this rewrite 🥳

Excellent!

Again, there is way too many things in this PR for me to be able to review it in details.

> * Value specifications are supported.

What's the status of higher-order functions expecting pure functional arguments (discussed in previous comments)?

>   The syntax is fairly different
>   however, we had discussed this a few months ago in the last type
>   checker meeting before the rewrite.  Basically, specifications are
>   written so that `requires`, `modifies`, `consumes`, `diverges` and `pure`
>   annotations are written first, then we have a `let` to bind the return
>   variables and the arguments.  This is then followed by a series of
>   `ensures` clauses.  In the case of exceptional post conditions, the
>   user can write `match ... with` and match over the corresponding
>   exceptional cases.

I see in the tests that this means that we are using names for the argument **before** they are bound by a `let`. This feels very unnatural. In the same spirit, this also means that the `match ... with` construct will be binding the names of the arguments.

> * When we state `modifies x` (or any ownership clause), we check if `x` is
>   a mutable type.  If not, we raise an error. A variable `x` is
>   considered mutable if it uses any type names that are mutable, with
>   the caveat that all arrow types are considered immutable.

This is nice. I'm not sure we often write a modify clause about an immutable value, but Gospel accepting `modify True` always felt a bit silly.

> Some relevant things that we still haven't done:
> 
> * Pattern matching.  We are trying to go without it and for the most
>   part it is doable, however we still have the issue of how to work
>   with the `option` type.

What's the rational for going without pattern matching? I remember saying there is something a bit strange about Gospel supporting partial function **and** providing (non-optional) pattern-matching exhaustivity. But honestly, I miss the ability to pattern match on the head of a sequence. I do believe that "You want to write some specification for your OCaml code? Use this language that looks like one of the ml family but without pattern matching!" will be a hard sell. That being said, there is no need to include that in the current PR.

Also, we are now supposed to write value specifications with a `match ... with`, does that mean this will only pattern match on value / exceptions or will we be able to destruct the returned value (its model)? I'd want to write:

```ocaml
val pop_opt : 'a queue -> 'a option
(*@ modifies q
    match pop_opt q with
     | None -> old q = q = empty
     | Some x -> q = cons x (old q) *)
```

In the meantime, we'll at least need a `fold` combinator on option type. (Which makes me think: what about `result`?)

> * Module types and Functors.  The former should be easy, the later
>   might be tricky.  I don't know if there are a lot of Ortac specs
>   that use functors, but if it's negligible, we could merge without
>   them.

Module types are currently supported. I believe at some point they should be re-include (again, in a follow up PR).

Current Gospel doesn't really support Functor. When you are in the body of the Functor, you don't have access to symbols from the module argument.

You have an example of how we deal with the kind of Functor accepted by Gospel in the `varray` example in [here](https://github.com/ocaml-gospel/ortac/tree/main/examples) (fair warning, it is kind of hackish).

> There's a lot more to this pull request, but they are either details or things that we have already discussed.

There are a **lot** of breaking changes. They should be summarized somewhere if we want to make this rewrite part of a new Gospel release.

> I wanted to take this opportunity to ask if you could send me the cleaned up tests you talked about a few weeks ago.

They are part of what has been removed by 456ac956

> Also, feel free to complain if you have any problems with the implementation/organisation.

- The gospel standard/logical library placed in the  `bin/` folder feels a bit weird. I think it would be reasonable for it to have its own folder at the root of the project.
- I think you can safely rename `src/tast2.ml` to `src/tast.ml` :-)

Thanks again. I'll start to familirise myself with the new code base and make make comments along the way.

mrjazzybread:
Hello

I changed the two things that you asked me to.

Regarding the absence of pattern matching (and ADTs) it is mostly to
normalise how users are supposed to use Gospel types. I also feel like
some form of pattern matching is going to be necessary, at the very
least we are going to need some convenient syntax to deal with ~option~
and, as you said, ~result~.

Regarding the new style of specifications, I feel like its pretty
similar to what would find in the literature, the only difference is
you don't have an explicit ~forall~ at the start explicitly binding each
name.  If it is just too weird, we will have to think of some
alternate syntax for nested exceptional specifications.

I have also added two .md files at the root of the directory, one with
breaking changes and other describing some of the new features that I
added.  The other points that you brought up in your last post are
addressed there.

mrjazzybread:
Hello

Just a quick post saying that I made a few updates to the type checker,
mostly just changing how the code is laid out.

n-osborne:
Thanks for the heads-up :-)

I'm wondering: would it be a good time to try to put the PR in a stable state (CI green), merge it somewhere (either `main` or a new `trunk` branch if you prefer).
I feel like you are now tackling some new features compared to the general rewrite (*e.g* making `let`s destruct tuples). Smaller and targeted PRs for these new features would allow me to review them.

mrjazzybread:
Hello

Regarding lets destructing tuples, I only did that because, since we
do not have pattern matching anymore, we had no way of doing anything
with tuples.  This is the last fully fledged feature that I am going
to add for this PR.

I am right now writing a pretty printer for the untyped AST, something
that always annoyed me about the previous type checker was that the
untyped AST did not have a good pretty printer so it was sometimes
difficult to figure out why things weren't working.  After I have that
and a pretty printer for the typed AST, I think we should have a type
checker meeting and (hopefully) merge this onto `main`.  This process
shouldn't take very long (I hope to get it done today).

mrjazzybread:
OK, after a bit more work, I believe I have done all I wanted with
this rewrite.  Besides all of the things that I have previously
mentioned, I have added:

-   A pretty printer for the typed AST

-   Added a function to produce typed ASTs from a list of file names.
    This is useful for verification tools that interface with Gospel
    since it is now more convenient to type check Gospel files.  A
    consequence of this is that I had to create two packages in the
    `dune-project` file:
    
    -   `Gospel_checker` - like the name suggests contains the
        utilities necessary to type check Gospel specifications
    -   `Gospel` - contains `Gospel_checker` as well as well as a utility
        function that takes a set of Gospel files and type checks them.
    
    The reason we need these two packages is because we use the
    utilities in `Gospel_checker` to create the files with the marshalled
    representations of the namespace for the Gospel standard library and
    the environment of OCaml primitives.  Since the aforementioned
    utility function depends on these files, it has to be placed in a
    separate compilation unit.

-   Top level structures such as functions, predicates and OCaml values
    now have associated with them the type variables that they use.  In
    other words, they are parameterised by their type variables.  This
    includes not only type variables named by the user, but also any
    type variable introduced by Inferno.  For example:
    
        (*@ function empty : 'a sequence *)
        
        (*@ axiom empty_length : length empty = 0 *)
    
    Here, the `empty` function uses the type variable `'a`, so the list of
    type variables associated with it will be `['a]`.  Additionally, when
    we type check the `empty_length` axiom we introduce a fresh type
    variable `'a`, meaning that, although the axiom does not use any named
    type variables, the list of type variables will be `['a]`.  Although
    this is probably not useful for Ortac, it greatly simplifies
    translations into Roqc

-   Related to the previous point, I also added a new constructor to
    typed terms `Ttyapply`, which receives an identifier for a top level
    variable and the types it receives as argument.  In the previous
    example, `empty` would be represented as `Ttyapply(empty, ['a])`.  Once
    again, Ortac can just probably just treat this constructor the same
    way it does the `Tvar` constructor, but this also helps with
    translations into Rocq.

With all of this, I think I am satisfied with the current state of
this PR.  I think that we are at a point where we could benefit from
a type checker meeting.  When would it be convenient for you?

mrjazzybread:
Ok, the CI is better now.  It technically failed, but that was just the
experimentals (and some weird internal errors which I think were
github's fault).

PR #451, opened 2025-09-15
Removed GitHub workflow files for documentation

mrjazzybread:
Hello

Small PR to remove the workflow files generating the documentation.  In the future, these should be replaced with ones that generate a website using Odoc.


n-osborne:
Thanks!
This makes sense and is obviously correct :-)

CI  failure is unrelated, merging.

PR #452, opened 2025-09-16
Value headers now appear at top of the spec

mrjazzybread:
Hello

Another small PR, this one is to update the parser with the syntax we discussed in the previous online meeting.

Right now, the Gospel we just merged onto main has a syntax where
headers appear in the middle of the specifications.  This PR changes
this so that they always appear at the top.  Additionally there is
still syntax for nested exceptional specifications.  As a small
example:

    val f : ref int -> unit
    (*@ f x
            consumes x
            raises E3
              produces x
              ensures x > 0
            raises E2 x
              ensures x = 0 *)

In this specification, the specification for `E3` is delimited by the
next `raises` clause.  These nested specifications only allow `produces`
and `ensures` clauses and they can be used to talk about different
ownership for different exceptions.  For example, in this
specification, when `E3` is raised, the user retains ownership of `x` but
they do not when `E2` is raised

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,6 +31,8 @@
   [\#404](https://github.com/ocaml-gospel/gospel/pull/404)
 - Changed the gospel typechecker to use bool as the type of logical formulae
   [\#391](https://github.com/ocaml-gospel/gospel/pull/391)
+- Add nested exceptional specifications
+  [#452](https://github.com/ocaml-gospel/gospel/pull/452)
 
 ## Internals
 


n-osborne:
Thanks!

I haven't read the code, but I have a few questions:

- Are the indentation meaningful?
- Do we still have the `checks` clause?
- Can we write `raises E0 | E1` and `raises _`?
- IIUC, we have to write **all** the specifications for the normal behaviour before the first `raises` clause?

mrjazzybread:
Hello

> Are the indentation meaningful?

No, they are just there so that the separation is more apparent.  As
with the rest of the Gospel language, users are free to indent their
specifications however they please.

> Do we still have the checks clause?

No, but that is an oversight.  I will add another commit for this
since I know current `Ortac` requires it.

> Can we write raises E0 | E1

No, but maybe it should be allowed.  Could also be interesting to
combine this with nested specifications, although this would be
non-trivial since it would require some changes in the typechecker.

> IIUC, we have to write all the specifications for the normal behavior
> before the first raises clause?

Yes, something that I kept from the previous parser that I liked is to
enforce an order for specifications: `requires`, `consumes`, `preserves`,
`modifies`, `pure`, `diverges` and `checks` must come before `ensures` and
`produces` clauses.  Additionally, as you said, these must all come
before `raises` clauses.

n-osborne:
>  >    Can we write raises E0 | E1

> No, but maybe it should be allowed. Could also be interesting to
> combine this with nested specifications, although this would be
> non-trivial since it would require some changes in the typechecker.

There is no need to over-complicates things.

> Yes, something that I kept from the previous parser that I liked is to
> enforce an order for specifications: requires, consumes, preserves,
> modifies, pure, diverges and checks must come before ensures and
> produces clauses. Additionally, as you said, these must all come
> before raises clauses.

IIRC, gospel 0.3 doesn't enforce an order in the clauses. But I do think that enforcing the order is a good thing.

n-osborne:
Just saw your last edits about fixity of symbols.
I'm not sure I understand the relation with the topic of the PR.
In order to ease review so that we can merge faster, please try to keep PR focused on one topic.

That being said, I'm not very enthusiastic about encoding the fixity of a symbol as a prefix in the string representing its name.

mrjazzybread:
Yes, that was an accident, I meant to push that to another branch, I have taken care of that

> That being said, I'm not very enthusiastic about encoding the fixity of a symbol as a prefix in the string representing its name.

Yes, I agree, Gospel currently encodes addition as "infix +".  I never liked this but I never got around to changing this convention.  One of my next PRs is about removing this feature

mrjazzybread:
I have done everything I wanted to do with this PR.  Whenever ready, you may merge.

PR #453, opened 2025-09-30
Identifiers now contain field stating if they are infix

mrjazzybread:
Hello

This is the fixity PR we had previously discussed.  I had already completed it by the time you sent me yours but after reviewing it I think our solutions are similar.

### DIFF
--- a/src/ident.ml
+++ b/src/ident.ml
@@ -53,6 +53,7 @@ module IdTable = Hashtbl.Make (Tag)
 
 type t = {
   id_str : string;
+  id_fixity : Preid.fixity;
   id_attrs : string list;
   id_loc : Location.t;
   id_tag : Tag.t;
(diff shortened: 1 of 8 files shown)


n-osborne:
Thanks!

> I had already completed it by the time you sent me yours but after reviewing it I think our solutions are similar.

That's good sign.
Also a quick look to your PR seems to indicate that the modification is simpler after your rewrite (#449) than before (from what I remember anyway). That's very good news!

> Small heads up, I just pushed some small changes

Ping me when the PR is stable so that I can review it.

mrjazzybread:
It's already stable.

n-osborne:
Thanks.

Just took a quick look.

I just have a small remark:

`Preid.fixity` defines a `Let` variant that is not used anywhere (not sure what it correspond to neither).

Other than that, LGTM.

PR #455, opened 2025-10-28
Fix bug in building constraints for lambdas

mrjazzybread:
Hello

This is a small bug fix PR.  When type checking usages of anonymous functions,
Gospel would check if the body of the function was well typed, but it would not
check if the expected type was a function.  For instance:

```
fun (x -> x + 0 <-> x)
```
Would not type check since `x` is used as a `prop` and and an `integer`.

However, the following would type check:

```
(fun x -> x) + (fun x -> x)
```

Although `+` expects two integers.  This is because there was no constraint stating that 
the type of lambda expressions is a function type.


mrjazzybread:
Ok, I have updated the change log and the comment

PR #457, opened 2025-11-05
More expressive let pattern matching

mrjazzybread:
Hello

Just up front, this PR is not implementing what we discussed in the
previous Gospel meeting.  This is instead a small extension to the
existing `let` pattern matching that allows for more complex patterns.
I implemented this when I made the `let*` proposal but I think it makes
sense as its own PR.

Currently, we can destruct tuples as such in Gospel:

    (*@ axiom let_example : forall x. let y, z, w = x in ... *)

But we cannot write, for instance:

    (*@ axiom let_example : forall x. let y, (z, w) = x in ... *)

In other words, we cannot use Gospel a `let` to destruct nested tuples.
Additionally, we cannot write casts such as:

    (*@ axiom let_example :
          forall x.
          let (y : integer, z : integer) = x in ... *)

With this PR, the two previous examples are now valid in Gospel.  Like
in OCaml, tuples can be destructed at arbitrary depth and casts can
appear anywhere in the pattern.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -33,6 +33,8 @@
   [\#391](https://github.com/ocaml-gospel/gospel/pull/391)
 - Add nested exceptional specifications
   [#452](https://github.com/ocaml-gospel/gospel/pull/452)
+- More expressive let pattern matching
+  [#457](https://github.com/ocaml-gospel/gospel/pull/457)
 
 ## Internals
 


mrjazzybread:
OK good.  I have rewritten the commit message so there is nothing else I am going to add to this

PR #461, opened 2025-11-14
Add test for function returning an option

n-osborne:
Test is failing, complaining to don't know about `option` type constructor.


n-osborne:
CI failure unrelated. I'm merging.

Issue #462, opened 2025-11-14
`option` in OCaml signature raises Unbound type constructor error


n-osborne:
#461 introduces the corresponding test case.

PR #463, opened 2025-11-14
Add back type checking of README's example

n-osborne:
Test is failing related to #462


n-osborne:
Removed the unsupported (for now) bits from the README so it can be merged.

PR #468, opened 2025-11-26
Improve behaviour regarding unsupported constructs

### LINKED ISSUE #467: Making type checker ignore unsupported constructs
The current behaviour when the type checker encounters an unsupported construct
(extensible type, polymorphic variant...) is to fail with an `assert false`.

The behaviour of gospel.0.3.0 is to fail with a more informative error message.

This is easy to port to the dev version, but it is still a problematic
behaviour: the type checker fail even if the Gospel annotations don't talk
about the unsupported construct. As a consequence, the type checker will for
example fail on an interface containing a declaration of a pretty printer
(using `Format` module) (see #450).

Another example where this behaviour is problematic is the use of
`ppx_deriving_qcheck` (as detailed in this
[discussion](https://discuss.ocaml.org/t/parsing-interfaces-that-use-external-modules-with-gospel/17496)).

A better behaviour would be to ignore OCaml type declarations and value
descriptions using one of these unsupported constructs. This would allow the
Gospel type checker to check the annotations of the supported part of a module.

Although, in case some Gospel specifications are attached to the unsupported
declaration, I think we shouldn't silently ignore it. One could argue we should
fail in this case, or we could chose to just emit a warning that the
specifcation will be ignored (I'm in favour of the former).

n-osborne:
This patch proposes to improve how unsupported OCaml constructs are handled by the Gospel type-checker.

Fixes #467

The current behaviour is to fail with an `assert false` when we encounter an unsupported construct.
This is decided when parsing the OCaml attributes containing the Gospel annotation (with the AST traversal).

The proposed behaviour is to fail with an informative error message when an unsupported OCaml contruct is annotated with some Gospel specifications.
Unupported OCaml constructs that are not annotated are simply ignored.

The information carried by the error is the name of the construct and the location so that the incriminated line of code is displayed with the error message.

Caveat: `test/Typechecking/unsupported/variant_as_argument.mli` reads as follow:

```ocaml
type t = A | B

val f : t -> bool
(* {gospel_expected|
[1] File "variant_as_argument.mli", line 3, characters 8-9:
    3 | val f : t -> bool
                ^
    Error: Unbound type constructor t
    
|gospel_expected} *)
```

What is happening is that the type declaration is ignored (unsupported but not specified), the type-checker later adds a header to the unspecified value definition and fail to find `t` in the typing environment.

But this is related to another problem, hence out of scope of this patch.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,8 @@
 
 ## Improved
 
+- Improve how unsupported OCaml constructs are handled
+  [\#468](https://github.com/ocaml-gospel/gospel/pull/468)
 - Fix filename for `check_file` output
   [\#478](https://github.com/ocaml-gospel/gospel/pull/478)
 - Add `option` as an OCaml primitive type constructor
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,8 @@
 
 ## Improved
 
+- Improve how unsupported OCaml constructs are handled
+  [\#468](https://github.com/ocaml-gospel/gospel/pull/468)
 - Fix filename for `check_file` output
   [\#478](https://github.com/ocaml-gospel/gospel/pull/478)
 - Add `option` as an OCaml primitive type constructor
--- a/src/uattr2spec.ml
+++ b/src/uattr2spec.ml
@@ -74,6 +74,8 @@ let rec preid_of_long (loc : location) (s : longident) =
   | Ldot (id, s) -> Qdot (preid_of_long id, Preid.create ~loc s)
   | _ -> assert false
 
+exception Unsupported
+
 let rec core_to_pty cty =
   let loc = cty.ptyp_loc in
   match cty.ptyp_desc with
@@ -82,7 +84,7 @@ let rec core_to_pty cty =
       PTtyapp (preid_of_long id.loc id.txt, List.map core_to_pty l)
   | Ptyp_arrow (_, t1, t2) -> PTarrow (core_to_pty t1, core_to_pty t2)
   | Ptyp_tuple l -> PTtuple (List.map core_to_pty l)
-  | _ -> assert false (* TODO replace with unsupported*)
+  | _ -> raise Unsupported
 
 let ptype_kind = function
   | Ptype_abstract -> PTtype_abstract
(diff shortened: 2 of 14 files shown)


mrjazzybread:
Hello

This is a very sensible way of dealing with the problem of unsupported
constructs.  Regarding your caveat, Gospel doesn't actually generate
headers for values without specifications.  Rather, it generates
headers for specifications that do not have one.  For instance:

    val neg : int -> int
    (*@ pure *)

The reason for that error is that Gospel does name resolution on the
type of OCaml values even if they do not have a specification attached
to them.  To prevent this, I wrote a small patch and pushed it to my
fork <https://github.com/mrjazzybread/gospel/tree/pr>.  Some of the
tests had to be changed so that OCaml values that are used in
specifications have a dummy Gospel specification to signal that they
are relevant for other specifications.  I think that this patch should
be added in this PR once it is merged.

There is also something strange happening with the tests, for some
reason when it prints the name of the file in the error it now starts
with "/." whereas it previously didn't do this.  This is likely some
package that was recently updated but I have no idea which one.  The
patch also deals with this.

n-osborne:
The problem in the tests was due to dune and is fixed in #494

n-osborne:
> (...) Some of the
> tests had to be changed so that OCaml values that are used in
> specifications have a dummy Gospel specification to signal that they
> are relevant for other specifications.

I don't think this is a good idea on the design perspective: we are asking the user to do something specific (and not really meaningful) just for the sake of our implementation.

What do you think about managing a set of unsupported type constructor?
The idea is that when we encounter a non-annotated unsupported type declaration, we still ignore it but also put its identifier in an environment to remember it is unsupported so that when it appears in another item we know what to do (raise an error if is annotated and ignore it if not).

mrjazzybread:
I was going through the list of PRs and I realized missed your comment
:|

> I don't think this is a good idea on the design perspective: we are
> asking the user to do something specific (and not really meaningful)
> just for the sake of our implementation.

I don't think it is unreasonable and in fact it is how tools like
Verifast handle the same problem: if something is not annotated it is
ignored.  As an example:

    val x : 'a ref
    (*@ ensures True *)
    
    val f1 : 'a -> unit
    (*@ f1 y
        modifies x
        ensures x = y *)

In order to be able to type check the specification for `f1`, we
have to give a dummy annotation to the value `x` so that Gospel knows
that that it has to remember the value `x` for later.  I feel like this
is the most sensible way to do this, it is only intrusive in the edge
case where the user wants to use an OCaml value in another
specification but does not want to specify said value.

> What do you think about managing a set of unsupported type
> constructor?

I like this and have thought about implementing it but it felt like a
bit of work for something that wouldn't be super useful to me since
all it would enable are more clearer error messages.  This should
probably be in a future To-Do list but I don't promise that I would
tackle this in the near future.

n-osborne:
superseded by #498

PR #476, opened 2026-02-18
Special syntax for building sets

mrjazzybread:
Hello

This is another small PR.  This one is for adding special syntax to
define sets using propositions.  The syntax is:

    { v | P }

Where `v` is some arbitrary name and `P` is a proposition that uses `P`.
This creates a set where every `P` holds for every element `v`.  For
instance if we wanted to define the set of positive integers, we would
state:

    { n | n => 0 }

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## Added
 
-- Add test cases for some unsupported contructs
+- Add test cases for some unsupported constructs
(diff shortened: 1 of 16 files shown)


n-osborne (approved):
Thanks.
LGTM

May I ask what is the use case?
I believe Ortac won't be able to consume these new constructions.

PR #479, opened 2026-05-28
Add Option module to Gospel stdlib

### LINKED ISSUE #469: Add an `Option` module in Gospel logical library
I open this issue to start discussing what should be in an `Option` module in the Gospel logical library.

A minimal set of functions would be:

- `get : 'a option -> 'a` (partial function)
- `is_some : 'a option -> bool` (or `prop` ? I believe we will want a `bool` version, see #460)
- `is_none : 'a option -> bool` (same remark)

n-osborne:
This PR proposes to add

- `Option.is_none`
- `Option.is_some`
- `Option.get`

to the Gospel standard library.

Closes https://github.com/ocaml-gospel/gospel/issues/469

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 - Add failing test case about map
   [\#480](https://github.com/ocaml-gospel/gospel/pull/480)
+- Add `Option.{get,is_none,is_some}` in Gospel stdlib
+  [\#479](https://github.com/ocaml-gospel/gospel/pull/479)
 - Add test cases for some unsupported constructs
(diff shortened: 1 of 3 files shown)


mrjazzybread:
Hello (sorry for the obscenely late reply)

The PR looks good to merge.  There are a few function that might also make sense to add (e.g. `map`)
but we can add those when if/when we find the need for them.

n-osborne:
Thanks @mrjazzybread :-)

Indeed, the idea is to add the minimal set of functionalities so we can start talking about `option`s. We'll see later whether we need more vocabulary.

Rebased on #494 to fix tests.
Merging.

PR #480, opened 2026-05-28
Add failing test case about map

n-osborne:


n-osborne:
CI is taking an unreasonable amount of time to find workers for the two last remaining architecture (riscv).
There is enough green to merge.

PR #496, opened 2026-09-15
Packaging: fix sub packaging

### LINKED ISSUE #495: packaging: `opam install gospel` won't install the `gospel_checker` library
At the `dune` level, two packages are defined:

1. `gospel`
2. `gospel_checker`

The latter being a dependency of the former.

At the `opam` level only one package is defined: `gospel`.

This means that when we install gospel via `opam`, it won't build: `gospel_checker` is missing from `gospel` dependencies.

I'm not sure why you (@mrjazzybread) wanted to declare the contents of `src/` in a different package (and what is the meaning of `dev_tools/`).

The previous version had one package exposing an executable and a library. We can regain this state by moving `dev_tools/gospel.ml[i]` to `src/` and making it exposes the modules we want to expose.

If we decide to keep the idea of separation, we can define a sub-package (see [here](https://dune.readthedocs.io/en/stable/concepts/package-spec.html#libraries)).

n-osborne:
This patch proposes a fix for gospel packaging.

Fixes #495

We declare the library defined in `src/` as a dune sub-package called `gospel.typecheckerlibs` (on the model of ocaml `compilerslibs` for lack of better name).
We also rename the `dev_tools` directory as `frontend` as it contains the implementation of the frontend function for gospel type-checker.
The said function is also renamed from `tast` to `typecheck` for explicitness / aesthetic reasons.

### Reproduce the bug:

- create a brand new and shiny dune project and add gospel as a dependency (also in one of the dune rule, not just the `dune-project`)
- (optionally to avoid messing with one of your existing switches) create a local opam switch for the new project
- `opam pin` gospel on the current `main` branch (`a6589b30fa83da0950f937066c7e915ad9782dde`)
- `dune build` will fail with something like:

```sh
46 |  (requires gospel_checker ppx_blob)
                ^^^^^^^^^^^^^^
Error: Library "gospel_checker" not found.
-> required by library "gospel" in
   .../_opam/lib/gospel
-> required by executable main in bin/dune:3
-> required by _build/default/bin/main.exe
-> required by _build/install/default/bin/bob
-> required by _build/default/bob.install
-> required by alias all
-> required by alias default
```

### Test the fix:

- `opam pin` gospel on the current patch
- `dune build` should happily succeed


src/dune
```diff
@@ -7,7 +7,8 @@
   (-v --explain)))
 
 (library
- (public_name gospel_checker)
+ (name typecheckerlibs)
+ (public_name gospel.typecheckerlibs)
```
Julow, line 11: The public OCaml name is `Typecheckerlibs` which doesn't sound specific to gospel.
Do you expect a name such as `Gospel.Typecheckerlibs` ?

Julow (commented):
The packaging is right :)

src/dune
n-osborne: Thanks :-)

This library is intended for project manipulating the Gospel typed AST.
The library is used in OCaml code with an `open Typecheckerlibs`, which, I agree, is not very specific (in addition to be quite long and susceptible to typo).
I am trying to avoid the repetition in the sub-package name (as in for example `gospel.gospel-typechecker-libs`), thinking that context clarify things.
One could argue that a dependency is declared once in the dune file but may be opened more than once in OCaml files.

Julow: I mean that if you depend on `gospel`, you have the `Typecheckerlibs` module in your environment, which doesn't seem right.
Usually, it would be named `Gospel_typecheckerlibs` as the library is named `gospel.typecheckerlibs`. You can eventually alias it to `Gospel.Typecheckerlibs` to create a hierarchy in the modules that looks good in the doc.

n-osborne: Thanks for the suggestion. That makes sense :-)

n-osborne:
@mrjazzybread any insight on the packaging and / or naming?

mrjazzybread:
Hello

Thank you for cleaning up after me :) I will first explain why there
are these two packages since I have realized that this information is
not in the repo anywhere and it might not be entirely obvious.  One
annoying problem I kept running into with the old Gospel typechecker
was when making changes to the standard library.  If it did not type
check, Gospel would still compile but would complain when you tried to
type check anything.  Fixing the type checking error would also be
difficult due to the weird way the old Gospel type checker dealt with
the standard library.

Because of this, I wanted to have it so that when Gospel type checker
is compiled, the standard library is type checked and if there is an
error, compilation fails.  To do this, I need to run the Gospel type
checker on the standard library.  This creates a circular dependency
in that compiling the type checker depends on type checking the
standard library and type checking the standard library depends on
compiling the type checker.  My solution was to create a package with
the type checking logic (`gospel_checker`) that the `gospel` package would
use to type check the standard library.

Although I think that this architecture is correct, I knew that the
way I did it was error prone, but I never got around to fixing it
since it worked well enough for me.  I have tested this fix (as well
as checking if it still rejects an invalid gospel standard library)
and it seems good.  I like the names as well.  I didn't know that you
could create "sub-packages" within existing packages so this makes a
lot more sense.

n-osborne:
Thanks for your feedback @mrjazzybread 

I've learned about dune sub-package when working on [QCheck-STM and lin](https://github.com/ocaml-multicore/multicoretests). They are neat.

n-osborne:
I'll do a last clean-up (I see at least one typo in git messages) and then merge.

PR #498, opened 2026-09-23
Improve behaviour on unsupported OCaml subset (take 2)

### LINKED ISSUE #467: Making type checker ignore unsupported constructs

n-osborne:
This patch series proposes to improve how Gospel handle the unsupported OCaml subset.

This fixes #467

This is doing things quite differently, and hopefully in a better way, than #468.

The general idea is to:

- fail on the unsupported part of OCaml when it is annotated with Gospel specification
- silently ignore, but keep in the internal representations, if it is not annotated

The main design difference from #468 is to keep information in internal representations (ASTs and typing environment).
This allow in particular to spot named unsupported types occurring in value declaration and exception definition, and exceptions with unsupported types in Gospel annotations.

For example this non-annotated extensible type declaration:

```ocaml
type t = ..
```

will be ignored but kept in the internal representations, including the typing environment, so that the occurrence in

```ocaml
val f : t -> int
(*@ i = f t *)
```

will be spotted and type-checking will fail.

The patch series is organised as follow:

1. The two first commits are preparation (utils and tests)
2. The two following commits deal with collecting unsupported OCaml signature items syntactically recognisable (polymorphic variant declaration, first class module, object fragment of OCaml...)
3. The three following commits deal with collecting, in the typing environment, identifiers of unsupported items that can occurs in other places (types and exceptions)
4. The three next commits deal with failing properly at the right places

The unsupported items dealt with in `2` are spotted when parsing the Gospel attributes, in `Uattr2spec` module.
Those collected in the typing environment, in `3`, are spotted during the name resolution step, in the `Typing` module.

The idea of including the unsupported subset of OCaml in Gospel ASTs (in `Sig_unsupported` nodes) was at first to be able to pass those items from the parsing phase to the name resolution one. But we keep those node in the final type AST so that Ortac/Wrapper can respect its contract (producing a module with the same signature of the input module).


mrjazzybread:
Thank you for the patch!  I will be reviewing it over the weekend.

src/uattr2spec.ml
```diff
@@ -161,42 +185,47 @@ let sig_exception exn =
...
 *)
 let rec signature_item_desc ~filename = function
-  | Psig_value v ->
-      let v = val_description ~filename v in
-      Some (Sig_val v)
-  | Psig_type (_, tl) ->
-      let* tl = map_option (type_declaration ~filename) tl in
-      Sig_type tl
+  | Psig_value v as s -> (
+      try Sig_val (val_description ~filename v) with
```
Julow: The exception is caught in several places and this might be annoying over time (eg. what if `module_declaration` raises in the future). I suggest catching the exception in the parent once and for all.

src/uattr2spec.ml
```diff
@@ -74,17 +81,28 @@ let rec preid_of_long (loc : location) (s : longident) =
   | Ldot (id, s) -> Qdot (preid_of_long id, Preid.create ~loc s)
   | _ -> assert false
 
-let rec core_to_pty cty =
+let rec core_to_pty ~annotated cty =
```
Julow: `~annotated` is passed everywhere and adds a bit of complexity. It's only needed to render the error message ? What do you think of computing it at the time of catching the exception with a `is_annotated : signature_item -> bool` ?

src/uattr2spec.ml
```diff
@@ -161,42 +185,47 @@ let sig_exception exn =
     or is an OCaml value declaration without a specification, we return [None].
```
balat: is it still true?

src/namespace.mli
```diff
@@ -136,6 +138,10 @@ val get_exn_info :
 (** [get_exn_info defs id] receives an exception identifier [id] and the types
     of its arguments. *)
 
+val is_unsupported_ocaml : mod_defs -> Parse_uast.qualid -> bool
+(** [is_unsupported_ocaml_type defs id] decides whether the type name [id]
```
balat: [is_unsupported_ocaml] I guess

src/namespace.ml
```diff
@@ -508,6 +524,16 @@ let add_ocaml_val vid tvars vty defs =
 
 let add_ocaml_val env tvars id ty = add_def (add_ocaml_val tvars id ty) env
 
+let add_unsupported_ocaml id defs =
```
balat: What if there is the same name in unsupported_env and type_env?
For example if you define two things with the same name do you remove from the unsupported_env when you define the second?

